PR #1, opened 2015-07-24
fix 'Debian Template: qubes-core-agent.postinst causes very slow package installation

adrelanos:
https://github.com/QubesOS/qubes-issues/issues/1065


adrelanos:
No longer required. Done by:
https://github.com/marmarek/qubes-core-agent-linux/pull/11

PR #5, opened 2015-10-30
qfile-unpacker: Avoid data loss by checking for child errors

rustybird:
When qfile-unpacker's child encountered an error, it would display an
error message and exit(1), but the parent didn't inspect its status and
exited successfully.

That was unfortunate for qvm-move-to-vm: Even if the destination VM e.g.
didn't have enough free disk space, the RPC call would claim to succeed
anyway, so the file would be deleted from the source VM.


rustybird:
Replaced commit with more compact version, which preserves the child's exit status (instead of mapping all nonzero exit statues to 1)

rustybird:
... and move the two checks closer together

rustybird:
To explain it better, the bugfix is that `return ret`, which is equivalent to `return (ret & 0377)` in main, which is always 0 unless the child process was terminated by a signal, is changed in this commit to `return WEXITSTATUS(ret)`.

PR #6, opened 2015-11-18
qfile-unpacker: syncfs() to avoid qvm-move-to-vm data loss

rustybird:


marmarek:
There was fdatasync() call at each file, but was removed for performance reasons:
https://github.com/QubesOS/qubes-linux-utils/commit/c1d42f160299c0524b8d65115ca1bff6d521277a
Having syncfs after the whole transfer indeed is a good idea, but it should be made before sending final result (and CRC) - to ensure that remote end will _not_ receive transfer confirmation if anything goes wrong. Take a look at do_unpack function in above commit.

PR #9, opened 2016-02-10
Let ntpdate use the ntp servers received by dhcp options

Zrubi:
NetworkManager (by default) do not care about ntp servers received by dhcp options
This patch help to distribute the NTP servers - at least for ntpdate, which is used by qvm-sync-clock

### DIFF
--- a/network/qubes-nmhook
+++ b/network/qubes-nmhook
@@ -7,3 +7,10 @@ if [ -x /bin/systemctl ]; then
 else
     /usr/sbin/service qubes-updates-proxy try-restart
 fi
+
+# Let ntpdate use the ntp servers received by dhcp options:
+if [ -z "${DHCP4_NTP_SERVERS}" ]; then
+    echo "pool.ntp.org" > /etc/ntp/step-tickers
+else
+    echo ${DHCP4_NTP_SERVERS} |tr '[[:space:]]' '\n' > /etc/ntp/step-tickers
+fi


marmarek:
I don't think `pool.ntp.org` should be hardcoded here. Someone may want to configure other server by direct modification of `/etc/ntp/step-tickers`. Also it should be possible to opt-out of this feature - we have `/etc/qubes/protected-files.d` for that (list of files which shouldn't be automatically modified by Qubes scripts).

Fro `protected-files.d` usage example see here: https://github.com/QubesOS/qubes-core-agent-linux/blob/master/network/setup-ip#L58-L64

And IMO this script should backup original content somewhere (`/etc/ntp/step-tickers.qubes-orig`?) and use that backup to restore original content in case of no NTP servers from DHCP. Be careful to not override that backup with already modified file...

Zrubi:
Well You are right.

But because this is actually a missing feature from NetworkManager I just wanted to make it simple.
Will trying to improve it - but no idea how to solve he opt-out feaute. It should be implemented in NM (like the DNS settings, and routing) instead of dispatcher scripts.

marmarek:
So, maybe better approach would be sending a patch to NetworkManager?

marmarek:
Since in Qubes 4.0 we no longer use ntpdate (we rely on whatever time sync service is running in sys-net by default - systemd-timesyncd, ntpd etc), this is irrelevant now.

PR #14, opened 2016-05-29
Fallback to zenity if kdialog not available

unman:
The issue with dolphin identified in https://github.com/QubesOS/qubes-issues/issues/1429 arises because kdialog isn't installed in debian template. It can't be made a dependency of the core-agent package as it pulls a ridiculous amount of kde stuff with it.
This quick fix closes issue with dolphin.

### DIFF
--- a/qubes-rpc/qvm-copy-to-vm.kde
+++ b/qubes-rpc/qvm-copy-to-vm.kde
@@ -20,21 +20,35 @@
 #
 #
 
+if type kdialog 2> /dev/null; then
+kdlg=1
+else kdlg=0
+fi
+
+if [ $kdlg -eq 1 ] ; then
 VM=$(kdialog -inputbox "Enter the VM name to send files to:")
+else
+VM=$(zenity --entry --text="Enter the VM name to send files to:")
+fi
+
 if [ X$VM = X ] ; then exit 0 ; fi
 
 SIZE=$(du --apparent-size -c -- "$@" 2> /dev/null | tail -1 | cut -f 1)
+if [ $kdlg -eq 1 ] ;then
 REF=$(kdialog --progressbar "Copy progress")
+fi
 qdbus $REF org.freedesktop.DBus.Properties.Set "" maximum $SIZE
 
 export PROGRESS_TYPE=gui
 
 /usr/lib/qubes/qrexec-client-vm $VM qubes.Filecopy \
 	/usr/lib/qubes/qfile-agent "$@" |
+if [ $kdlg -eq 1 ] ;then
 (while read sentsize ; do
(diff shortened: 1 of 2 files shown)


qubes-rpc/qvm-copy-to-vm.kde
```diff
@@ -20,21 +20,35 @@
...
+VM=$(zenity --entry --text="Enter the VM name to send files to:")
+fi
+
 if [ X$VM = X ] ; then exit 0 ; fi
 
 SIZE=$(du --apparent-size -c -- "$@" 2> /dev/null | tail -1 | cut -f 1)
+if [ $kdlg -eq 1 ] ;then
 REF=$(kdialog --progressbar "Copy progress")
+fi
 qdbus $REF org.freedesktop.DBus.Properties.Set "" maximum $SIZE
```
marmarek: This line should also be inside of `if`, as it uses `$REF` set there.

qubes-rpc/qvm-copy-to-vm.kde
```diff
@@ -20,21 +20,35 @@
...
 export PROGRESS_TYPE=gui
 
 /usr/lib/qubes/qrexec-client-vm $VM qubes.Filecopy \
 	/usr/lib/qubes/qfile-agent "$@" |
+if [ $kdlg -eq 1 ] ;then
 (while read sentsize ; do
 	CURRSIZE=$(($sentsize/1024))
 	qdbus $REF  org.freedesktop.DBus.Properties.Set "" value $CURRSIZE
 done)
+fi
```
marmarek: Keep in mind the output of `qrexec-client-vm` above is piped here, so probably some `else` clause should be used. Otherwise `qfile-agent` will get SIGPIPE on reporting progress. Or simply set `PROGRESS_TYPE` to empty string.

qubes-rpc/qvm-copy-to-vm.kde
unman: @marmarek 
Yes. I did think I would show progress bar in both cases, so had else, but I couldnt get zenity --progress to do its stuff, so just dropped it completely. Should have tidied up properly. I'll do that now.

marmarek: For `zenity` it needs passing progress on its stdin, not via dbus. This is already done in `qvm-copy-to-vm.gnome`. Hmm, maybe for this reason it would be better to fallback to that script, instead of adding `zenity` support here?

unman: You mean test for kdialog, and if fail call the gnome version? That sounds good.

marmarek: Yes, exactly.

PR #15, opened 2016-05-30
Fall back to gnome utilities if kdialog not present

unman:
Closes https://github.com/QubesOS/qubes-issues/issues/1429

### DIFF
--- a/qubes-rpc/qvm-copy-to-vm.kde
+++ b/qubes-rpc/qvm-copy-to-vm.kde
@@ -19,25 +19,28 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #
 #
+if type kdialog 2> /dev/null; then
+    VM=$(kdialog -inputbox "Enter the VM name to send files to:")
+    if [ X$VM = X ] ; then exit 0 ; fi
 
-VM=$(kdialog -inputbox "Enter the VM name to send files to:")
-if [ X$VM = X ] ; then exit 0 ; fi
+    SIZE=$(du --apparent-size -c -- "$@" 2> /dev/null | tail -1 | cut -f 1)
+    REF=$(kdialog --progressbar "Copy progress")
+    qdbus $REF org.freedesktop.DBus.Properties.Set "" maximum $SIZE
 
-SIZE=$(du --apparent-size -c -- "$@" 2> /dev/null | tail -1 | cut -f 1)
-REF=$(kdialog --progressbar "Copy progress")
-qdbus $REF org.freedesktop.DBus.Properties.Set "" maximum $SIZE
+    export PROGRESS_TYPE=gui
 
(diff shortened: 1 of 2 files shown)


qubes-rpc/qvm-copy-to-vm.kde
```diff
@@ -19,25 +19,33 @@
...
-# if ! [ "x"$agentstatus = xDONE ] ; then
-# 	kdialog --sorry 'Abnormal file copy termination; see /var/log/qubes/qrexec.xid.log in dom0 for more details'
-# fi
+    qdbus $REF close
+    # we do not want a dozen error messages, do we
+    # if ! [ "x"$agentstatus = xDONE ] ; then
+    # 	kdialog --sorry 'Abnormal file copy termination; see /var/log/qubes/qrexec.xid.log in dom0 for more details'
+    # fi
+else
+    /usr/lib/qubes/qvm-copy-to-vm.gnome "$@" &
```
marmarek: Why in the background? This, among other things, hides exit code.

qubes-rpc/qvm-copy-to-vm.kde
```diff
@@ -19,25 +19,33 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #
 #
+if type kdialog 2> /dev/null; then
+    kdlg=1
+    else kdlg=0
+fi
```
marmarek: I think it would be much more readable if the whole fallback is done here (no `kdialog` - call gnome version, maybe even `exec` it), instead of setting a variable to just check it once, a line later.

qubes-rpc/qvm-copy-to-vm.kde
unman: It didn't seem to me that the exit code is being used. Maybe I should look again. Seemed somehow cleaner to hand off completely to gnome script.

qubes-rpc/qvm-copy-to-vm.kde
unman: Well I had it in mind that there might be an issue assuming that zenity is present, and there might be a need to do a series of tests.

qubes-rpc/qvm-copy-to-vm.kde
marmarek: Even if exit code is not useful (not really sure about that - but nautilus/dolphin _may_ use it to report possible failure), running it in background may result in unexpected results.

qubes-rpc/qvm-copy-to-vm.kde
marmarek: core-agent package depends on `zenity`, so it should be there...

marmarek:
(for the next time) take a look at https://www.qubes-os.org/doc/code-signing/
We're going to enforce this rule soon (will be announced).

PR #16, opened 2016-06-13
misc: add qvm-features-request

woju:
This tool is used to request features from template.

QubesOS/qubes-issues#1637


marmarek:
This should go into Qubes 4.0 branch (`core3-devel`), not 3.2 (`master`).
Also I see totally unrelated change (`$(BINDIR)`) in the same commit.

PR #17, opened 2016-06-30
Order network management units after network-pre.target

rustybird:
Network management software should order itself after network-pre.target
(man 7 systemd.special) so that other units can order themselves before
the *beginning* of network initialization. (qubes-misc-post too because
it calls setup-ip.)

Relevant for QubesOS/qubes-issues#2108


marmarek:
Ok, will merge it soon.

PR #18, opened 2016-07-01
dispVM fixes

rustybird:


vm-systemd/qubes-qrexec-agent.service
```diff
@@ -1,6 +1,6 @@
 [Unit]
 Description=Qubes remote exec agent
-After=qubes-dvm.service
+After=qubes-dvm.service xendriverdomain.service
```
marmarek: Is this really necessary? I think ordering `xendriverdomain` after `qubes-dvm` should be enough.

rustybird: @marmarek: Hmm, I always think of qrexec-agent having started as "now the VM is in a reasonable state for other VMs or dom0 to communicate with it", e.g. to attach a block device to it. Sort of like basic.target for Qubes VMs...

marmarek: This is what `qubes.WaitForSession` service is for. And it basically wait for X session to be fully started (but can be extended to other things if needed).

rustybird: But in a dispVM qubes.WaitForSession is run *before* the save file creation.

marmarek: Ah, right. Anyway, ordering xendriverdomain before qrexec shouldn't be necessary. Especially having  #1990 fixed. But it may make DispVM startup longer (less parallel services startup)

rustybird: Okay, removed the qrexec-agent ordering.

rustybird: Wait a second though, if your concern is delayed startup: xendriverdomain.service does not need anything besides what qubes-qrexec-agent.service already depends on. It's literally just the daemon forking off. (I could make it even slimmer by writing a tiny native systemd unit instead to replace that gross rc.d script.) So all the code from https://github.com/marmarek/qubes-vmm-xen/commit/8639b6986ff15aed18c5f47ec56328b4a66f5763 could be avoided?

rustybird:
> Wait a second though, if your concern is delayed startup: xendriverdomain.service does not need anything besides what qubes-qrexec-agent.service already depends on. It's literally just the daemon forking off. (I could make it even slimmer by writing a tiny native systemd unit to replace that gross rc.d script.) So all the code from marmarek/qubes-vmm-xen@8639b69 could be avoided?

Proposed commit series pushed. I will of course revert it to previous one (without the ordering and without the native xendriverdomain.service) if you prefer.

vm-systemd/qubes-qrexec-agent.service
marmarek: > So all the code from marmarek/qubes-vmm-xen@8639b69 could be avoided?

Not really, it is still needed, but having native xendriverdomain.service will make it easier.

marmarek:
I like native `xendriverdomain.service`, but it should go into vmm-xen repository.

rustybird:
> I like native xendriverdomain.service, but it should go into vmm-xen repository.

True. Removed the last commit here, will submit it there

marmarek:
> True. Removed the last commit here, will submit it there

Create it with a patch (just like xl-devd-trigger), to be also compatible with debian build (where you can have only one source tarball...).

rustybird:
Done: https://github.com/QubesOS/qubes-vmm-xen/pull/3
Also pushed another commit here that enables the xendriverdomain.service preset.

PR #19, opened 2016-10-11
Remove custom mount when starting cron, in favour of bind-dirs

unman:
Removing custom mount from /rw/cron should resolve QubesOS/qubes-issues#2326 and QubesOS/qubes-issues#2327


unman:
@marmarek 
Set up bind-dirs for /var/spool/cron
I had thought of setting requirements that the bind mounts be in place before starting the service, but decided against as some users might want to disable system persistence but still use cron while VM is running.

vm-systemd/cron.service.d/30_qubes.conf
```diff
@@ -1,3 +1,5 @@
 [Unit]
 ConditionPathExists=/var/run/qubes-service/crond
+
+[Service]
```
marmarek: Ordering should be in `[Unit]` section.

version
```diff
@@ -1 +1 @@
-3.2.11
+3.2.12
```
marmarek: Please don't increment version, especially not in the same commit as other changes.

vm-systemd/crond.service.d/30_qubes.conf
```diff
@@ -1,3 +1,5 @@
 [Unit]
 ConditionPathExists=/var/run/qubes-service/crond
+
+[Service]
```
marmarek: Same here.

marmarek:
And also `debian/changelog`...

unman:
@marmarek  Finally. I wont do that again.

PR #20, opened 2016-10-12
Eliminate race condition with qubes-setup-dnat-to-ns

Rudd-O:
`qubes-setup-dnat-to-ns` is called multiple times during boot.  Of particular interest are the two invocations done by:

1. `/usr/lib/qubes/init/network-proxy.setup.sh` (`qubes-network.service`)
2. `/usr/lib/qubes/init/misc-post.sh` (`qubes-misc-post.service`)

These can, and do often, run in parallel.  Often enough that the `PR-QBS` `nat` chain can end up with eight rules instead of four, or (worse) zero rules.

This commit represents the proper boot ordering of these services, where the post startup *must* happen after Qubes has already started its iptables, firewall, network setup and netwatcher.

This eliminates the race.

I considered wrapping the execution of `qubes-setup-dnat-to-ns` in a `flock` exclusive lock to prevent the race, and that may still be a good idea to do, but the actual bug is the systemd boot race, which this commit fixes.

I suspect there's races in many places, we are just lucky we do not hit them, but in all honesty, all firewall manipulation should grab an exclusive lock, or else racing `iptables-save` and `iptables-restore` will only cause pain.

### DIFF
--- a/vm-systemd/qubes-misc-post.service
+++ b/vm-systemd/qubes-misc-post.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Qubes misc post-boot actions
-After=network-pre.target qubes-dvm.service qubes-mount-dirs.service
+After=network-pre.target qubes-dvm.service qubes-mount-dirs.service qubes-network.service qubes-firewall.service qubes-netwatcher.service
 
 [Service]
 Type=oneshot


marmarek:
I think this may fix https://github.com/QubesOS/qubes-issues/issues/1067. Is it possible?

Rudd-O:
I do not know if that bug will be fixed, but it is certainly possible if said bug is caused by a race.

Rudd-O:
Also, PLEASE MERGE! :-)

Rudd-O:
For the love of god, can you please release this?  Qubes OS networking is literally unusable without this PR.  Half the time I boot I have to keep rebooting ProxyVMs just so I can get DNS resolution.

marmarek:
It was in current-testing for past month. Now moved to current. I forgot to place 'Fixes xxx' in commit, so automatic notifications are missing.

PR #24, opened 2016-11-13
Initialize home_volatile for disposable VMs.

Rudd-O:
When DVM savefile creation process starts, we initialize the DVM /home_volatile/user with the files from /etc/skel or /home.orig/user (depending on which one exists) unconditionally.

### DIFF
--- a/init/functions
+++ b/init/functions
@@ -115,3 +115,78 @@ umount_retry() {
     done
     return 0
 }
+
+initialize_home() {
+    local home_root
+    local mode
+    local user
+    local uid
+    local gid
+    local homedir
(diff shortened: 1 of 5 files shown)


Rudd-O:
Should address https://github.com/QubesOS/qubes-issues/issues/2390#issuecomment-259288205.

marmarek:
Looks ok, thanks!
Will merge it as soon as I can test it out (which may be near end of the week...).

Rudd-O:
Did it work as advertised?

marmarek:
Yes, thanks! :)

PR #25, opened 2016-11-14
Add systemd override for haveged in xenial and stretch. 

unman:
QubesOS/qubes-issues#2161 - 
Issue arises in both stretch and xenial. Solution is to provide override file, changing ordering of startup.
Reenable haveged.service after Debian package installation, to force new link in multi-user.target.wants after upgrade.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -263,6 +263,11 @@ else
 	install -m 0644 misc/py2/xdg.py* $(DESTDIR)/$(PYTHON_SITEARCH)/qubes/
 endif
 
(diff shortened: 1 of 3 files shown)


unman:
I don't think you can change WantedBy in the Install section in a conf file, and we *do* want to change that. That's why we override the whole file.

marmarek:
Ah, indeed, from systemd.unit manual:

    Also note that settings from the "[Install]" section are not honoured in drop-in unit files, and have no
    effect.

Ok then.

PR #30, opened 2016-12-06
misc: add qvm-features-request

woju:
This tool is used to request features from template.

QubesOS/qubes-issues#1637


woju:
second attempt, previous commit also did unrelated changes to `Makefile`

PR #31, opened 2016-12-13
Introduce $QREXEC_REMOTE_DOMAIN_*ID* in VMs

rustybird:
Surface the numerical Xen domid of the qrexec connection's remote VM to the local program as `$QREXEC_REMOTE_DOMAIN_ID`.

Use case: This variable makes it surprisingly easy to get client VMs to communicate with sys-whonix via Socks-over-qrexec, while preserving Tor circuit isolation between the different client VMs:

    #!/bin/sh
    #
    # /etc/qubes-rpc/SocksAdapter in sys-whonix

    n=$QREXEC_REMOTE_DOMAIN_ID
    src_ip=127.137.$((n / 256)).$((n % 256))

    exec socat STDIN TCP4:127.0.0.1:9150,bind=$src_ip

(The client VM would run the opposite adapter, using `socat` or a systemd socket unit to redirect Socks requests from e.g. the browser to `qrexec-client sys-whonix SocksAdapter`.)

### DIFF
--- a/qrexec/qrexec-agent-data.c
+++ b/qrexec/qrexec-agent-data.c
@@ -57,13 +57,16 @@ static void sigusr1_handler(int __attribute__((__unused__))x)
     signal(SIGUSR1, SIG_IGN);
 }
 
-void prepare_child_env() {
+void prepare_child_env(int remote_domid) {
     char pid_s[10];
+    char remote_domid_s[16];
 
     signal(SIGCHLD, sigchld_handler);
     signal(SIGUSR1, sigusr1_handler);
     snprintf(pid_s, sizeof(pid_s), "%d", getpid());
     setenv("QREXEC_AGENT_PID", pid_s, 1);
+    snprintf(remote_domid_s, sizeof(remote_domid_s), "%d", remote_domid);
+    setenv("QREXEC_REMOTE_DOMAIN_ID", remote_domid_s, 1);
 }
 
 int handle_handshake(libvchan_t *ctrl)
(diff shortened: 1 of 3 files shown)


marmarek:
Coincidence with IP address being generated from QID is just a
implementation detail and should not be relied on. In fact it will
change Qubes 4.0:
 - https://github.com/QubesOS/qubes-issues/issues/1143
 - https://github.com/QubesOS/qubes-issues/issues/1477

In this case you can achieve the same using source VM name as socks
authentication (on ports with IsolateSOCKSAuth).

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

rustybird:
In this case it doesn't matter whether the qid (or rather, Xen domid) is related to the usual IP address or not. Any kind of short enough (more than 16 bits is inconvenient) trusted number unique to a VM while it is running will work. The number is stuffed into an 127.* IP address, so it will always differ from the usual 10.* or whatever address, and a VM's Socks-over-qrexec streams will then *also* be isolated from the same VM's Socks-over-TCP streams. But that's totally fine.

The nice thing about doing VM circuit isolation this way is that the raw unmodified Socks streams can be passed from the client VM's programs to the tor daemon. Possibly with user and password already set, e.g. by Tor Browser for further isolation based on the URL bar domain and the current "New Tor Circuit for this site" nonce.

marmarek:
Still, VM ID should be considered an implementation detail, not part of public API. There are a lot of situations where we may want to change it - for example we may want to route vchan connection through some other VM (for example to hide real remote domain ID, or implement some extended auditing).
If you want numeric identifier, get some hash of domain name, or generate it randomly and build a map.

rustybird:
Fair enough. The second approach (keeping a map) turned out less gnarly than I expected.

PR #33, opened 2017-01-14
Shut down after update only if it's a template.

lorenzog:
As per discussion in
https://github.com/QubesOS/qubes-issues/issues/2555#issuecomment-271415169

Note: I've kept the semicolon at the end of the command, but I'm not sure why it was there in the first place. Tested on Debian Standalone VM and Template.

### DIFF
--- a/qubes-rpc/qubes.InstallUpdatesGUI
+++ b/qubes-rpc/qubes.InstallUpdatesGUI
@@ -17,7 +17,7 @@ elif [ -e /etc/arch-release ]; then
 else
     update_cmd='echo Unsupported distribution, install updates manually; bash -i'
 fi
-xterm -title update -e su -l -c "$update_cmd; echo Done.; echo Press Enter to shutdown the template, or Ctrl-C to just close this window; read x && poweroff;"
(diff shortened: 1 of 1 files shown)


marmarek:
Please sign your commit - see https://www.qubes-os.org/doc/code-signing/

marmarek:
Besides missing signature, it's ok.

lorenzog:
@marmarek I've generated a key and signed, but the automated check says unable to verify (no valid key found). I've then uploaded the key to keys.gnupg.net.. does it need to re-run or am I missing something?

lorenzog:
The key is also in my github setting and on pool.sks-keyservers.net

marmarek:
I don't see the key on keys.gnupg.net:
```
gpg --keyserver keys.gnupg.net --recv 572149EC63E49216
```
But it is on the other keyserver, so it's ok now.

marmarek:
The script checking this isn't automatically rerun, I'll trigger it in a moment.

lorenzog:
Thanks. I tried closing and re-opening but didn't help. I'll upload the signing key too.

PR #34, opened 2017-02-03
Add archlinux support to upgrade checker

torrancew:
This commit adds Archlinux support to the upgrade checker utility, via `checkupdates`, which is part of the `pacman` package:

```
$ pacman -Qo `which checkupdates`
/usr/bin/checkupdates is owned by pacman 5.0.1-4
```


torrancew:
I will get git and GPG wired up this weekend and sign this change.

marmarek:
Ping?

na--:
I'll fix this later. At a glance, there is a slight problem with it anyway - I think `grep '->'` should be `grep '\->'` or better yet: `grep -F -- '->'`. With the current code there is an error `grep: invalid option -- '>'`.

marmarek:
Superseded by #59

PR #35, opened 2017-02-06
Apply constraints to anacron as well as cron in Debian and Ubuntu TemplateBasedVMs

unman:
Closes QubesOS/qubes-issues#2109


marmarek:
Thanks, generally it looks good. Will look into details later, currently we're busy preparing GSoC application.

PR #36, opened 2017-02-08
add comments for Qubes onion apt repository

adrelanos:


marmarek:
Generally I'm ok with this, but please sign the commit (or push signed tag for it).

adrelanos:
So no more quick and easy github web changes. :|

Actually I don't like the style myself in this file too much. Let's see if there are better suggestions in https://github.com/QubesOS/qubes-issues/issues/2623.

PR #37, opened 2017-02-09
Fix build for trusty - locales-all not available

unman:
One of fixes for QubesOS/qubes-issues#2608

### DIFF
--- a/Makefile.builder
+++ b/Makefile.builder
@@ -10,6 +10,10 @@ ifeq ($(PACKAGE_SET),vm)
   ARCH_BUILD_DIRS := archlinux
 endif
 
+ifeq ($(DIST),trusty)
+  $(shell sed -i /locales-all,/d  $(CHROOT_DIR)/$(DIST_SRC)/debian/control)
+endif
+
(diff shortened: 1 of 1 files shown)


unman:
@marmarek Should I submit separate PRs against release3.2, as well as master, or will you handle this?

marmarek:
`master` only is ok, I'll handle migrating it to `release3.2`.

marmarek:
This is wrong place - `$(shell ...)` will be executed every time `Makefile.builder` gets parsed. You should place it in `source-debian-quilt-copy-in` target, and without `$(shell ...)`.

Makefile.builder
```diff
@@ -10,6 +10,10 @@ ifeq ($(PACKAGE_SET),vm)
   ARCH_BUILD_DIRS := archlinux
 endif
 
+ifeq ($(DIST),trusty)
+  $(shell sed -i /locales-all,/d  $(CHROOT_DIR)/$(DIST_SRC)/debian/control)
+endif
+
```
marmarek: This is wrong place - `$(shell ...)` will be executed every time `Makefile.builder` gets parsed. You should place it in `source-debian-quilt-copy-in` target, and without `$(shell ...)`.

PR #38, opened 2017-02-11
Reset iptables ACCEPT rule for updates proxy if service is running

unman:
Closes QubesOS/qubes-issues#1555

### DIFF
--- a/network/qubes-firewall
+++ b/network/qubes-firewall
@@ -51,6 +51,10 @@ while true; do
 		DISPLAY=:0 /usr/bin/notify-send -t 3000 "Firewall loading error ($(hostname))" "$OUT" || :
 	fi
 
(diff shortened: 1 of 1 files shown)


adrelanos:
Backticks...
```
`...`
```
are the legacy syntax. Can you please move to `$(...)`?

As per http://mywiki.wooledge.org/BashFAQ/082.

unman:
On Sun, Feb 12, 2017 at 04:22:22PM -0800, Patrick Schleizer wrote:
> Backticks...
> ```
> `...`
> ```
> are the legacy syntax. Can you please move to `$(...)`?
> 
> As per http://mywiki.wooledge.org/BashFAQ/082.
> 

It's neither obsolete or deprecated, and in this case perfectly
serviceable.

PR #39, opened 2017-02-12
Apply gschema override preventing previews in nautilus in Debian

unman:
QubesOS/qubes-issues#1108
The prevention of previews is already set in Fedora packages.
This applies it to Debian packages too.


adrelanos:
Will cause a lintian warning (at least in stretch).

    command-with-path-in-maintainer-script /usr/bin/glib-compile-schemas

Could you make this please...?

    /usr/bin/glib-compile-schemas -> glib-compile-schemas

marmarek:
Done.

adrelanos:
Cold you please add `|| true` or `&> /dev/null || true` at the end of `glib-compile-schemas`?

/var/lib/dpkg/info/libglib2.0-0:amd64.postinst uses `|| true`.

/var/lib/dpkg/info/qubes-gui-agent.postinst uses

    glib-compile-schemas /usr/share/glib-2.0/schemas &> /dev/null || true

PR #42, opened 2017-03-16
bind-dirs: Create ro if bind target exists

ypid:
Before, the script skipped none existing ro paths even if the path existed below /rw. This would require someone who wants to use bind-dirs only in TemplateBasedVM to ensure that the paths exist before the bind-dirs script gets called.

This patch changes this behavior so that if the path exists below /rw, it is ensured that an (empty) file/directory is present in ro (where the corresponding path from /rw is then bind mounted over).

Requires: Docs update. I can open a PR when this PRs looks good.
Fixes limitation: "Does not work if the file / folder in question does not already exist in the root image. I.e. a file that does not exist in the root image cannot be bind mounted in the TemplateBasedVM."
Example use case: https://github.com/debops/ansible-persistent_paths
Tested on: Qubes OS 3.2; Debian 8 TemplateBasedVM (and Template)

> > What's the workaround?
>
> Put the files in /rw/bind-dirs...
> Edit bind-dirs.sh in template to create the ro file/directory if it doesn't already exist. I wouldn't want to see this in the released code.

Ref: https://github.com/QubesOS/qubes-issues/issues/2661#issuecomment-284040319

Ping: @adrelanos, @unman, @Rudd-O, @marmarek
Does this work for you?

### DIFF
--- a/vm-systemd/bind-dirs.sh
+++ b/vm-systemd/bind-dirs.sh
@@ -87,15 +87,21 @@ bind_dirs() {
          continue
       fi
 
-      # Initially copy over data directories to /rw if rw directory does not exist.
-      if [ -d "$fso_ro" ] || [ -f "$fso_ro" ]; then
-         if ! [ -d "$fso_rw" -o -f "$fso_rw" ]; then
-            echo "Initializing $rw_dest_dir with files from $fso_ro" >&2
-            cp --archive --recursive --parents "$fso_ro" "$rw_dest_dir"
+      if [ -d "$fso_rw" ] || [ -f "$fso_rw" ]; then
+         if [ ! -e "$fso_ro" ]; then
+            ## Create empty file or directory if path exists in /rw to allow to bind mount none existing files/dirs.
+            test -d "$fso_rw" && mkdir --parents "$fso_ro"
+            test -f "$fso_rw" && touch "$fso_ro"
          fi
       else
-         true "$fso_ro is neither a directory nor a file or does not exist, skipping."
-         continue
+         if [ -d "$fso_ro" ] || [ -f "$fso_ro" ]; then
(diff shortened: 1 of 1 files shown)


vm-systemd/bind-dirs.sh
```diff
@@ -87,15 +87,21 @@ bind_dirs() {
          continue
       fi
 
-      # Initially copy over data directories to /rw if rw directory does not exist.
-      if [ -d "$fso_ro" ] || [ -f "$fso_ro" ]; then
-         if ! [ -d "$fso_rw" -o -f "$fso_rw" ]; then
```
adrelanos: Glad you rewrote that previously not very easily readable line.

adrelanos:
Thank you for improving bind-dirs!

Would you mind to check in further changes if this breaks Whonix?

----

a)

* See if a newly created whonix-gw based ProxyVM is still functional.
* Reboot it. See if it's still fine.

----

b)

* Check an existing sys-whonix (that was created prior the bind-dirs changes).
* `mount > mounts_before`
* `sudo journalctl -u qubes-mount-dirs -> qubes-mount-dirs_before`
* Restart an existing sys-whonix.
* Apply bind-dirs changes.
* `mount > mounts_after`
* `sudo journalctl -u qubes-mount-dirs -> qubes-mount-dirs_after`
* diff those files (probably using meld or kdiff3)
* If the bind-dirs related stuff does not differ, it should be fine.

----

Just did that with the changes from #42 and #43 at once. Doesn't break Whonix.

adrelanos:
Haven't tested the fixed (space handling) and new functionality (the fixed limitation).

ypid:
> Would you mind to check in further changes if this breaks Whonix?

Sure, I am going to do this in the future. Thanks for testing and the instructions!

PR #47, opened 2017-07-06
clock synchronization rewrite

marmarta:
clock synchronization mechanism rewritten to use systemd-timesync instead of NtpDate; at the moment, requires:
- modifying /etc/qubes-rpc/policy/qubes.GetDate to redirect GetDate to designated clockvm
- enabling clocksync service in clockvm ( qvm-features clockvm-name service/clocksync true )

Works as specified in issue listed below, except for:
- each VM synces with clockvm after boot and every 6h
- clockvm synces time with the Internet using systemd-timesync
- dom0 synces itself with clockvm every 1h (using cron)

fixes QubesOS/qubes-issues#1230


qubes-rpc/qubes-sync-clock
```diff
@@ -0,0 +1,38 @@
...
+#
+
+import sys
+import re
+import subprocess
+
+def main():
+    stdin = sys.stdin.read(25)
+
+    date_out = stdin.strip()
```
kalkin: I think this should be called untrusted_date_out to highlight that the data is untrusted. Shouldn't it? @marmarek

marmarek:
> I'm not sure if syncing dom0 & vm's every X hours is useful and if it creates privacy issues.

There is no ntp running in every VM (intentionally, to not expose it to ntp parsing bugs), so some time synchronization is needed. On Qubes 3.x it's done every 6 minutes, which is way too often.
On the other hand, this approach allows you to selectively disable it using qrexec policy.

qubes-rpc/qubes-sync-clock
```diff
@@ -25,12 +25,12 @@ import subprocess
...
 
-    date_out = stdin.strip()
+    untrusted_date_out = stdin.strip()
 
-    if not re.match(r'^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+00:?00$', date_out):
+    if not re.match(r'^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+00:?00$', untrusted_date_out):
         sys.stderr.write('Invalid date received, aborting!')
         sys.exit(1)
-    subprocess.check_call(['date', '-u', '-Iseconds', '-s', date_out],
+    subprocess.check_call(['date', '-u', '-Iseconds', '-s', untrusted_date_out],
```
marmarek: After validation of the value, drop `untrusted_` prefix, see here: https://www.qubes-os.org/doc/coding-style/#security-coding-guidelines

PR #55, opened 2017-09-23
dnf-qubes-hooks: handle newer DNF >= 2.x

fepitre:
dnf-qubes-hooks: handle newer DNF >= 2.x

### DIFF
--- a/misc/dnf-qubes-hooks.py
+++ b/misc/dnf-qubes-hooks.py
@@ -20,8 +20,10 @@
 #
 
 from __future__ import absolute_import
+from distutils.version import LooseVersion
 import logging
 import dnf
+import dnf.const
 import subprocess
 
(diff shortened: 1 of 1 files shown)


misc/dnf-qubes-hooks.py
```diff
@@ -35,7 +35,10 @@ def __init__(self, base, cli):
         self.log = logging.getLogger('dnf')
 
     def transaction(self):
-        config = self.read_config(self.base.conf, PLUGIN_CONF)
+        try:
+            config = self.read_config(self.base.conf, PLUGIN_CONF)
+        except:
```
marmarek: This will catch any exception, not only related to wrong parameters. Ideally, there should be some condition checking dnf version. But if not possible, at least add `TypeError` here.

PR #57, opened 2017-09-29
Fix the archlinux package - DROPIN dirs, shellcheck issues, indentation

na--:
Without this change the package builds successfully but there is a file conflict error when installing it. This is because the default value for `SYSTEM_DROPIN_DIR` in the [Makefile](https://github.com/QubesOS/qubes-core-agent-linux/blob/a7ef5726ed368b97acb2f6497dece22ca9f9fdd3/Makefile#L55) is `lib/systemd/system`, however `/lib` is a symlink to `/usr/lib` in archlinux. I do not know how the [package](http://olivier.medoc.free.fr/archlinux/current/qubes-vm-core-3.2.18-11-x86_64.pkg.tar.xz) in the [repository](https://github.com/QubesOS/qubes-core-agent-linux/blob/a7ef5726ed368b97acb2f6497dece22ca9f9fdd3/archlinux/PKGBUILD#L92-L94) is not broken. When building this myself, there's a `lib` folder in the package root that conflicts with the symlink.

Also, I am not sure what the purpose of the service `getty@tty1.service.d` is ([here](https://github.com/QubesOS/qubes-core-agent-linux/blob/a7ef5726ed368b97acb2f6497dece22ca9f9fdd3/vm-systemd/getty@tty.service.d/30_qubes.conf), [here](https://github.com/QubesOS/qubes-core-agent-linux/blob/a7ef5726ed368b97acb2f6497dece22ca9f9fdd3/archlinux/PKGBUILD#L96-L102) and [here](https://github.com/QubesOS/qubes-core-agent-linux/blob/a7ef5726ed368b97acb2f6497dece22ca9f9fdd3/archlinux/PKGBUILD.install#L359-L361)). I've had it disabled for months with no issues.
 
And a last remark (that probably should be a separate  issue) - [shellcheck](https://www.shellcheck.net/) lights up like a christmas tree when I run it on most of the scripts. Most of the errors are relatively harmless ones like [SC2006](https://github.com/koalaman/shellcheck/wiki/SC2006) and [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086) (unless we have spaces in the filenames), but still.

### DIFF
--- a/archlinux/PKGBUILD
+++ b/archlinux/PKGBUILD
@@ -57,14 +57,14 @@ build() {
     sed 's:/usr/sbin/qubes-firewall:/usr/bin/qubes-firewall:g' -i vm-systemd/qubes-firewall.service
 
     for dir in qubes-rpc qrexec misc; do
-      (cd $dir || exit 1; make)
+        make -C "$dir"
     done
 }
 
 package() {
     # Note: Archlinux removed use of directory such as /sbin /bin /usr/sbin (https://mailman.archlinux.org/pipermail/arch-dev-public/2012-March/022625.html)
     # shellcheck disable=SC2154
-    (cd qrexec || exit 1; make install DESTDIR="$pkgdir" SBINDIR=/usr/bin LIBDIR=/usr/lib SYSLIBDIR=/usr/lib)
+    make -C qrexec install DESTDIR="$pkgdir" SBINDIR=/usr/bin LIBDIR=/usr/lib SYSLIBDIR=/usr/lib
 
     make install-vm DESTDIR="$pkgdir" SBINDIR=/usr/bin LIBDIR=/usr/lib SYSLIBDIR=/usr/lib SYSTEM_DROPIN_DIR=/usr/lib/systemd/system USER_DROPIN_DIR=/usr/lib/systemd/user DIST=archlinux
 
--- a/archlinux/PKGBUILD
+++ b/archlinux/PKGBUILD
@@ -1,11 +1,8 @@
-# This is an example PKGBUILD file. Use this as a start to creating your own,
-# and remove these comments. For more information, see 'man PKGBUILD'.
-# NOTE: Please fill out the license field for your package! If it is unknown,
-# then please put 'unknown'.
-
+#!/bin/bash
 # Maintainer: Olivier Medoc <o_medoc@yahoo.fr>
+# shellcheck disable=SC2034
 pkgname=qubes-vm-core
-pkgver=`cat version`
+pkgver=$(cat version)
 pkgrel=11
 epoch=
 pkgdesc="The Qubes core files for installation inside a Qubes VM."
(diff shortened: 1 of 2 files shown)


marmarek:
> Also, I am not sure what the purpose of the service getty@tty1.service.d

Well, the first file you've pointed out disable it, unless you specifically enable it with `qvm-service --enable ... getty@tty`.

> And a last remark (that probably should be a separate issue) - shellcheck lights up like a christmas tree when I run it on most of the scripts.

It wasn't that bad: https://github.com/QubesOS/qubes-core-agent-linux/pull/58
Ok, it was. Anyway, fixed: https://travis-ci.org/marmarek/qubes-core-agent-linux/jobs/281551625

na--:
That was fast! :) I will fix the shellcheck warnings in `PKGBUILD` and `PKGBUILD.install` files later today as well and push them in this pull request.
Regarding `getty@tty1.service.d` - doesn't [this](https://github.com/QubesOS/qubes-core-agent-linux/blob/a7ef5726ed368b97acb2f6497dece22ca9f9fdd3/archlinux/PKGBUILD.install#L359-L361) enable it after every install/update of the package? My archlinux template is a bit customized and it's difficult to test this properly, but since I'll be building a fresh template today to debug some pvgrub issues anyway, I'll test this as well.

marmarek:
> Regarding getty@tty1.service.d - doesn't this enable it after every install/update of the package?

No, Condition* in systemd units are considered at runtime - when starting the service. So service can be enabled, but will not be started unless you create that file (which is done by qvm-service tool).

na--:
Sorry for the delay, debugging the pvgrub issues took a while yesterday. Fortunately [the problem](https://bugs.archlinux.org/task/55447) was in the Xen or the kernel, not in QubesOS. Recompiling the kernel (the major timesink) with ` CONFIG_INTEL_ATOMISP=n` or using the `linux-lts` kernel fixes the pvgrub boot. Not sure if this will be an issue with Fedora/Debian pvgrub at some point or if it's Archlinux specific.

Since I was fixing shellcheck related things all over `PKGBUILD` and `PKGBUILD.install` I decided to fix the indentation as well. The default github diff is not pretty, but [hiding the whitespace differences](https://github.com/QubesOS/qubes-core-agent-linux/pull/57/files?w=1) makes it somewhat readable.

And you are right, no issues with `getty@tty1.service.d`, or at least not when you leave it disabled as it is by default. I remember manually disabling it a while ago, after I noticed that it was constantly spamming errors in the logs. I had never enabled it manually or via `qvm-service`, so I don't know why it was started, but that's fixed now. Still, if I enable it manually now wih `systemctl start getty@tty1.service.d`, I get these errors in the logs every 10 seconds:
```
Oct 01 22:09:04 test systemd[1]: getty@tty1.service.d.service: Service has no hold-off time, scheduling restart.
Oct 01 22:09:04 test systemd[1]: Stopped Getty on tty1.service.d.
Oct 01 22:09:04 test systemd[1]: Started Getty on tty1.service.d.
Oct 01 22:09:04 test agetty[2686]: /dev/tty1.service.d: cannot open as standard input: No such file or directory
Oct 01 22:09:14 test systemd[1]: getty@tty1.service.d.service: Service has no hold-off time, scheduling restart.
Oct 01 22:09:14 test systemd[1]: Stopped Getty on tty1.service.d.
Oct 01 22:09:14 test systemd[1]: Started Getty on tty1.service.d.
Oct 01 22:09:14 test agetty[2687]: /dev/tty1.service.d: cannot open as standard input: No such file or directory
```
Still not sure what the service is supposed to do, but since it's disabled by default I do not care enough to debug it. 

And lastly, I [disabled](https://github.com/QubesOS/qubes-core-agent-linux/pull/57/commits/5662d7e5fe7f5236a2623f725b7e0f908d26631f) Oliver's binary pacman repo by default because it broke the build with an error `qubes-r3.2: key "2043E7ACC1833B9C" is unknown` and also to harmonize it with the [current documentation](https://www.qubes-os.org/doc/templates/archlinux/#binary-packages-activation).

marmarek:
cc @ptitdoc

na--:
Sorry, I think I tested an old build before pushing the previous commit and missed the breakage. At least the old adage "if it ain't broke, don't fix it" was proven once again... :)

marmarek:
I see why you've disabled binary repository, but that means users will not get updates out of the box. Better to fix packages available in @ptitdoc repository...

archlinux/PKGBUILD
```diff
@@ -25,87 +22,84 @@ options=()
...
+    sed 's:#!/usr/bin/python:#!/usr/bin/python2:' -i qubes-rpc/*
+    sed 's:#!/usr/bin/env python:#!/usr/bin/env python2:' -i qubes-rpc/*
+
+    # Fix for archlinux sbindir
+    sed 's:/usr/sbin/ntpdate:/usr/bin/ntpdate:g' -i qubes-rpc/sync-ntp-clock
+    sed 's:/usr/sbin/qubes-netwatcher:/usr/bin/qubes-netwatcher:g' -i vm-systemd/qubes-netwatcher.service
+    sed 's:/usr/sbin/qubes-firewall:/usr/bin/qubes-firewall:g' -i vm-systemd/qubes-firewall.service
+
+    for dir in qubes-rpc qrexec misc; do
+      (cd $dir || exit 1; make)
```
marmarek: make -C $dir

archlinux/PKGBUILD
```diff
@@ -25,87 +22,84 @@ options=()
...
+    done
 }
 
 package() {
-  # Note: Archlinux removed use of directory such as /sbin /bin /usr/sbin (https://mailman.archlinux.org/pipermail/arch-dev-public/2012-March/022625.html)
-
-  (cd qrexec; make install DESTDIR=$pkgdir SBINDIR=/usr/bin LIBDIR=/usr/lib SYSLIBDIR=/usr/lib)
+    # Note: Archlinux removed use of directory such as /sbin /bin /usr/sbin (https://mailman.archlinux.org/pipermail/arch-dev-public/2012-March/022625.html)
+    # shellcheck disable=SC2154
+    (cd qrexec || exit 1; make install DESTDIR="$pkgdir" SBINDIR=/usr/bin LIBDIR=/usr/lib SYSLIBDIR=/usr/lib)
```
marmarek: make -C qrexec ...

archlinux/PKGBUILD.install
```diff
@@ -64,89 +62,89 @@ configure_notification-daemon() {
...
+    #  echo >> /etc/yum.conf
+    #  echo '# Yum does not support inclusion of config dir...' >> /etc/yum.conf
+    #  echo 'include=file:///etc/yum.conf.d/qubes-proxy.conf' >> /etc/yum.conf
+    #fi
+    #/usr/lib/qubes/update-proxy-configs
+    # Archlinux pacman configuration is handled in update_finalize
+
+    # Location of files which contains list of protected files
+    mkdir -p /etc/qubes/protected-files.d
+    # shellcheck source=../init/functions
```
marmarek: shellcheck is called from source top dir and it expect paths relative from there...

archlinux/PKGBUILD.install
```diff
@@ -208,16 +206,16 @@ preset_units() {
...
                 systemctl --no-reload preset "$unit_name" >/dev/null 2>&1 || :
             fi
         fi
-    done
+    done < "$1"
 }
 
 restore_units() {
-    grep '^[[:space:]]*[^#;]' "$1" | while read action unit_name
+    while read -r action unit_name < grep '^[[:space:]]*[^#;]' "$1"
```
marmarek: This looks wrong

na--:
Thanks for the review, did not know about `make -C`! And not sure why I even changed `restore_units` or didn't fix it last night when I fixed basically the same mistake I made in the previous function, sorry...

Regarding the binary pacman repo - I personally prefer not to use it or to manually enable it if I have to. I may be a bit paranoid, but prefer not to trust such non-official builds and build the packages myself. Even non-official builds from people who basically wrote most of the arch code... actually, are @ptitdoc and Olivier Medoc the same person? Still, I don't mind enabling the unofficial repo by default and manually disabling it only for myself, I decided to just disable it because the official documentation said it shloud be disabled. I'll fix it if I can just figure out how to prevent the build from breaking with `qubes-r3.2: key "2043E7ACC1833B9C" is unknown`, any help is appreciated.

marmarek:
Regarding the binary repository - in fact builder-archlinux disable it just after installation, and indeed we have concluded that clear instruction how to enable them (if one want to) is enough. So this change is ok.

Which also remind me it would be good to build new template - the one in repository is almost year old.

marmarek:
One more thing: I'm going to put this to both master and release3.2 branches. Please make pull requests to master branch in the future (and state in description whether it's applicable for 3.2 too).

ptitdoc:
About SYSTEM_DROPIN_DIR, it is an uncommited change, my bad.

marmarek:
On Tue, Oct 03, 2017 at 03:17:52PM +0000, ptitdoc wrote:
> About SYSTEM_DROPIN_DIR, it is an uncommited change, my bad.

Tip: `make check` from time to time.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

ptitdoc:
Regarding `getty@tty1.service.d`, from what I remember, it was a related to broken access controls related to logind, pam.d when using su.

I had to change `pam.d/su` and `pam.d/su-l` to start a logind session as su is used by qrexec. However, I experienced a lot of access denied errors in polkit/dbus which where causing several things such as gvfs to fail. For this to work, I had to ensure that a user was explicitly logged in using the following tip https://wiki.archlinux.org/index.php?title=Automatic_login_to_virtual_console&redirect=no

ptitdoc:
Regarding the binary repository, I think that I also experienced build failure:
- Before building the gui-agent, the core-agent is installed in building chroot
- As soon as the core-agent is installed, the binary repository is enabled
- As soon as pacman -Suy is started (before building any qubes package), the error `qubes-r3.2: key "2043E7ACC1833B9C" is unknown` appears because it tries to download this repository in addition to the local build repository

The simple solution is probably to create a .disabled file for the binary repository from the start, and somehow make a reference to the documentation in this file.

na--:
@marmarek: do you want me to submit a cleaned-up (squashed) version of this pull request to the master branch as well or will you cherry pick? As far as I can see, there weren't a lot of differences in the archlinux folder before I made the shellcheck changes:
`git diff master..f7bf46f780047fbc57638207a9d250bad530cdcb -- archlinux/`
results in:
```diff
diff --git a/archlinux/PKGBUILD b/archlinux/PKGBUILD
index a239ff8..8ee625a 100644
--- a/archlinux/PKGBUILD
+++ b/archlinux/PKGBUILD
@@ -13,7 +13,7 @@ arch=("x86_64")
 url="http://qubes-os.org/"
 license=('GPL')
 groups=()
-depends=("qubes-vm-utils>=3.1.3" python2 python3 python2-xdg ethtool ntp net-tools gnome-packagekit imagemagick fakeroot notification-daemon dconf zenity qubes-libvchan "qubes-db-vm>=3.2.1" haveged python2-gobject python2-dbus xdg-utils notification-daemon gawk sed procps-ng librsvg)
+depends=("qubes-vm-utils>=3.1.3" python2 python3 python2-xdg ethtool ntp net-tools gnome-packagekit imagemagick fakeroot notification-daemon dconf pygtk zenity qubes-libvchan "qubes-db-vm>=3.2.1" haveged python2-gobject python2-dbus xdg-utils notification-daemon gawk sed procps-ng librsvg)
 makedepends=(gcc make pkg-config "qubes-vm-utils>=3.1.3" qubes-libvchan qubes-db-vm qubes-vm-xen libx11 python2 python3 lsb-release)
 checkdepends=()
 optdepends=(gnome-keyring gnome-settings-daemon networkmanager iptables tinyproxy python2-nautilus gpk-update-viewer)
@@ -56,6 +56,7 @@ sed 's:#!/usr/bin/env python:#!/usr/bin/env python2:' -i qubes-rpc/*
 
 # Fix for archlinux sbindir
 sed 's:/usr/sbin/ntpdate:/usr/bin/ntpdate:g' -i qubes-rpc/sync-ntp-clock
+sed 's:/usr/sbin/qubes-netwatcher:/usr/bin/qubes-netwatcher:g' -i vm-systemd/qubes-netwatcher.service
 sed 's:/usr/sbin/qubes-firewall:/usr/bin/qubes-firewall:g' -i vm-systemd/qubes-firewall.service
 
 for dir in qubes-rpc qrexec misc; do
diff --git a/archlinux/PKGBUILD.install b/archlinux/PKGBUILD.install
index 51823a7..7bb92bd 100644
--- a/archlinux/PKGBUILD.install
+++ b/archlinux/PKGBUILD.install
@@ -475,7 +475,7 @@ post_remove() {
 
     rm -rf /var/lib/qubes/xdg
 
-    for srv in qubes-sysinit qubes-misc-post qubes-mount-dirs qubes-network qubes-qrexec-agent; do
+    for srv in qubes-dvm qubes-sysinit qubes-misc-post qubes-mount-dirs qubes-netwatcher qubes-network qubes-qrexec-agent; do
         systemctl disable $srv.service
     done   
```

The problem is that I would be unable to test the changes, I'm still running QubesOS 3.2.

ptitdoc:
Same here. I don't have a test environment for QubesOS 4.0 yet. Should we blindly cherry-pick all commits util somebody can test the archlinux template with QubesOS 4.0 ?

marmarek:
> Same here. I don't have a test environment for QubesOS 4.0 yet. Should we blindly cherry-pick all commits util somebody can test the archlinux template with QubesOS 4.0 ?

I'd say yes, otherwise almost for sure we'll miss some commits later. Generally there is not so much difference in VM agents in 4.0 vs 3.2.

marmarek:
> or will you cherry pick?

Already done (you can see it in `master-staging` branch).

PR #59, opened 2017-10-04
Add archlinux support to upgrade checker (resubmit)

na--:
Since I don't think it's possible to edit another user's pull request, this PR includes the cherry-picked commit from the original [pull request](https://github.com/QubesOS/qubes-core-agent-linux/pull/34) by @torrancew and a small fix.

### DIFF
--- a/misc/upgrades-installed-check
+++ b/misc/upgrades-installed-check
@@ -23,6 +23,13 @@ elif [ -e /etc/debian_version ]; then
    apt_get_upgrade_output="$(LANG="C" apt-get -s upgrade 2>&1)"
    exit_code="$?"
    echo "$apt_get_upgrade_output" | awk "/^Inst/{ print $2 }" | [ "$(wc -L)" -eq 0 ] && echo "true" || echo "false"
+elif [ -e /etc/arch-release ]; then
+    ## Archlinux
+    set -e
+    set -o pipefail
+    checkupdates_output="$(checkupdates 2>&1)"
+    exit_code="$?"
+    echo "$checkupdates_output" | grep '->' | [ "$(wc -L)" -eq 0 ] && echo "true" || echo "false"
 else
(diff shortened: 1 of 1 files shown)


na--:
Forgot to mention that this should be applicable for both version 4.0 and version 3.2 (tested on 3.2).

misc/upgrades-installed-check
```diff
@@ -23,6 +23,13 @@ elif [ -e /etc/debian_version ]; then
    apt_get_upgrade_output="$(LANG="C" apt-get -s upgrade 2>&1)"
    exit_code="$?"
    echo "$apt_get_upgrade_output" | awk "/^Inst/{ print $2 }" | [ "$(wc -L)" -eq 0 ] && echo "true" || echo "false"
+elif [ -e /etc/arch-release ]; then
+    ## Archlinux
+    set -e
+    set -o pipefail
+    checkupdates_output="$(checkupdates 2>&1)"
+    exit_code="$?"
+    echo "$checkupdates_output" | { grep -F -- '->' || true; } | [ "$(wc -L)" -eq "0" ] && echo "true" || echo "false"
```
marmarek: wc seems useless here, grep itself can set appropriate exit code:
```
! echo "$checkupdates_output" | grep -qF -- '->' && echo "true" || echo "false"
```
(note `!` for negation)

na--: Do we need the negation? I like using grep's exit code directly, it greatly simplifies the expression, but can't we just switch true and false like this?
```
echo "$checkupdates_output" | grep -qF -- '->' && echo "false" || echo "true"
```
If there are updates, grep will match the arrows, exit with 0 and will output "false" (i.e. "there are pending upgrades"), otherwise it will output true ("all upgrades have been installed").

marmarek: Yes, that's even better.

PR #60, opened 2017-10-18
Archlinux template fixes for QubesOS 4.0

na--:
This fixes the archlinux `core-agent-linux` build for QubesOS 4.0. The binary pacman repo is completely disabled because as far as I know @ptitdoc builds only 3.2 packages there.

Parent issue: QubesOS/qubes-issues#3185

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -173,7 +173,11 @@ install-common: install-doc
 	# force /usr/bin before /bin to have /usr/bin/python instead of /bin/python
 	PATH="/usr/bin:$(PATH)" python setup.py install $(PYTHON_PREFIX_ARG) -O1 --root $(DESTDIR)
 	mkdir -p $(DESTDIR)$(SBINDIR)
+
+ifneq ($(SBINDIR),/usr/bin)
 	mv $(DESTDIR)/usr/bin/qubes-firewall $(DESTDIR)$(SBINDIR)/qubes-firewall
+endif
+
 
 	install -d -m 0750 $(DESTDIR)/etc/sudoers.d/
 	install -D -m 0440 misc/qubes.sudoers $(DESTDIR)/etc/sudoers.d/qubes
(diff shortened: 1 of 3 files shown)


ptitdoc:
Discussion related to disabling the binary repository here:
https://github.com/QubesOS/qubes-core-agent-linux/pull/57

ptitdoc:
Quoting @marmarek :

> Regarding the binary repository - in fact builder-archlinux disable it just after installation, and indeed we have concluded that clear instruction how to enable them (if one want to) is enough. So this change is ok.
> 
> Which also remind me it would be good to build new template - the one in repository is almost year old.

Quoting @ptitdoc:
> The simple solution is probably to create a .disabled file for the binary repository from the start, and somehow make a reference to the documentation in this file.

ptitdoc:
By the way, I can already build binaries for Qubes 4.0 if it is stable enough. I'm just not able to test it yet.

marmarek:
Lets keep .disabled file with repository definition.

na--:
I'll change it back and add a `PKGBUILD-qubes-repo-4.0.conf` file. 

@ptitdoc: what should the `Server` URL be? Since archlinux does not have the concept of distribution versions, I think that a separate repository address is needed. The 3.2 is `http://olivier.medoc.free.fr/archlinux/current`, how about `http://olivier.medoc.free.fr/archlinux/4.0`? This should be more maintainable in the long term. Also, I think that I can safely delete `PKGBUILD-qubes-repo-3.1.conf` since version 3.1 has been officially unsupported for half an year.

ptitdoc:
You can keep exactly the same URL.

Only replace to repository name from
`[qubes-r3.2]`
to
`[qubes-r4.0]`

On my side, I use the same directory, however, the pakage database qubes-r3.2.db and qubes-r4.0.db will reference different packages.

You can safely remove 3.1 data as it is now unmaintained. The file wil be anyway still present in the 3.1 branch.

Thanks for your help by the way.

ptitdoc:
I will try to build the modules for 4.0 the coming week.

na--:
This is superseded by https://github.com/QubesOS/qubes-core-agent-linux/pull/63, closing

PR #63, opened 2017-10-23
Archlinux r4 template

ptitdoc:
- Picked @na-- commits
- Reviewed changes since 3.2 release
- Changed pam.d logic from using su to using qrexec


ptitdoc:
These last commits should fix the qubes.PostInstall bug.

PR #66, opened 2017-10-29
Makefile split of core-agent and networking tools

ptitdoc:
Split install targets of the Makefile to add:
install-network: basic networking for VM
install-networkmanager: qubes network manager files
install-netvm: files required for a working netvm (include firewall stuff, dns ...

The main goal is to use it to generate different packages in archlinux, which requires files to be explicitly installed.

The Makefile could be splitted in more target as it could be used as documentation (role of the different files).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,13 +56,16 @@ SYSTEM_DROPIN_DIR ?= "lib/systemd/system"
 USER_DROPIN_DIR ?= "usr/lib/systemd/user"
 
 SYSTEM_DROPINS := chronyd.service crond.service cups.service cups.path cups.socket ModemManager.service
-SYSTEM_DROPINS += NetworkManager.service NetworkManager-wait-online.service getty@tty.service
-SYSTEM_DROPINS += tinyproxy.service
+SYSTEM_DROPINS += getty@tty.service
+
 SYSTEM_DROPINS += tmp.mount
 SYSTEM_DROPINS += org.cups.cupsd.service org.cups.cupsd.path org.cups.cupsd.socket
 SYSTEM_DROPINS += systemd-random-seed.service
 SYSTEM_DROPINS += tor.service tor@default.service
 SYSTEM_DROPINS += systemd-timesyncd.service
+
+SYSTEM_DROPINS_NETWORKING := NetworkManager.service NetworkManager-wait-online.service
+SYSTEM_DROPINS_NETWORKING += tinyproxy.service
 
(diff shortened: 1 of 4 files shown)


marmarek:
As in other repository: how do you envision handling upgrade? Will user be required to install new (sub-)package manually? If that's a common practice in Archlinux, then ok. But if there are some additional guidelines how to handle such cases (package split, package rename), then better to apply them here. Or maybe pacman is smart enough to handle it automatically?
FWIW for other distros we have meta-packages to ease handle such situations: https://github.com/qubesos/qubes-meta-packages

marmarek:
Ping @ptitdoc @na--

na--:
There are [package groups and meta-packages](https://wiki.archlinux.org/index.php/creating_packages#Meta_packages_and_groups) in archlinux as well, though I'm unsure how they can help with this package split. In any case, I don't think that the package split is much of a concern for QubesOS 4.0 (this pull request), since there's no automatic upgrade from 3.2 and users have to reinstall anyway. This would be more problematic for the [pull request to 3.2](https://github.com/QubesOS/qubes-core-agent-linux/pull/68/files), where an upgrade of `qubes-vm-core` may lead to loss of functionality if someone was using an arch appvm as a netvm.

I am not sure that there's an idiomatic way to handle such package splits in arch. Do you think that adding `qubes-vm-networking` to `qubes-vm-core`'s optional dependencies and/or a post-install/post-upgrade message will be enough?

ptitdoc:
`qubes-vm-networking` contains netvm related functions. I actually planned to make it optional in `qubes-vm-core` but I may have forgotten to to so.

An optional dependency in archlinux only means that you will have a message when installing the package, and that you will see the optional dependency when querying package information (`pacman -Qi quebs-vm-core` would show `qubes-vm-networking` as an optional dependency).

Of course we should document that option in Qubes-OS wiki, and include the package in full featured templates.

ptitdoc:
To summarize:

The package will be noted as Optional and documented in the wiki. We could of course include `qubes-vm-networking` in a full featured archlinux template.

ptitdoc:
I added optional dependencies in the most featured pull request of this branch:
https://github.com/QubesOS/qubes-core-agent-linux/pull/67/commits/5b45cf1808cb53c1f9e25cf6b90e09d6dbf02071

Sorry for the mess between the different pull requests. This one superseed all the qubes-vm-core pull requests:
https://github.com/QubesOS/qubes-core-agent-linux/pull/67

PR #67, opened 2017-10-31
Archlinux keyring

ptitdoc:
Attempt to implement binary repository keyring in archlinux so that users can activate the repository in a single command, without any internet access requirement.


ptitdoc:
Following discussions in https://github.com/QubesOS/qubes-core-agent-linux/pull/66, the Makefile split seems to be accepted however there are discussions on dependencies between `qubes-vm-core`, `qubes-vm-networking` and `qubes-vm-keyring`.

* `qubes-vm-core` is sufficient for a standard Qubes VM usage and include basic networking setup.
* `qubes-vm-networking` is required if a user want to use the archlinux template as a NetVM/ProxyVM. It is now an optional dependency of `qubes-vm-core`
* `qubes-vm-keyring` is required if a user want to enable the QubesOS binary packages maintained by Olivier Medoc. It is now an optional dependency of `qubes-vm-core`

archlinux/PKGBUILD-networking.install
```diff
@@ -0,0 +1,41 @@
...
+
+    # Yum proxy configuration is fedora specific
+    #if ! grep -q '/etc/yum\.conf\.d/qubes-proxy\.conf' /etc/yum.conf; then
+    #  echo >> /etc/yum.conf
+    #  echo '# Yum does not support inclusion of config dir...' >> /etc/yum.conf
+    #  echo 'include=file:///etc/yum.conf.d/qubes-proxy.conf' >> /etc/yum.conf
+    #fi
+
+    for srv in qubes-firewall.service qubes-iptables.service qubes-network.service qubes-updates-proxy.service ; do
+        systemctl enable $srv.service
```
marmarek: `.service` is added twice here.

archlinux/PKGBUILD-networking.install
```diff
@@ -0,0 +1,41 @@
...
+## arg 1:  the new package version
+## arg 2:  the old package version
+post_upgrade() {
+    post_install
+}
+
+## arg 1:  the old package version
+post_remove() {
+    for srv in qubes-firewall.service qubes-iptables.service qubes-network.service qubes-updates-proxy.service ; do
+        systemctl disable $srv.service
```
marmarek: `.service` is added twice here.

Makefile
```diff
@@ -336,4 +384,8 @@ install-deb: install-common install-systemd install-systemd-dropins
 	mkdir -p $(DESTDIR)/etc/systemd/system/
 	install -m 0644 vm-systemd/haveged.service  $(DESTDIR)/etc/systemd/system/
 
-install-vm: install-rh install-common
+install-corevm: install-rh install-common install-systemd install-sysvinit install-systemd-dropins install-networking
+
+install-netvm: install-systemd-networking-dropins install-networkmanager install-netvm
```
marmarek: `install-netvm` shouldn't be as dependency of itself...

ptitdoc:
Thanks a lot for the review.

Makefile
ptitdoc: Commited

archlinux/PKGBUILD-networking.install
ptitdoc: Commited

archlinux/PKGBUILD-networking.install
ptitdoc: Commited

PR #68, opened 2017-11-01
archlinux: pick Qubes4.0 mergeable changes

ptitdoc:


archlinux/PKGBUILD
```diff
@@ -1,19 +1,17 @@
 #!/bin/bash
 # Maintainer: Olivier Medoc <o_medoc@yahoo.fr>
 # shellcheck disable=SC2034
-pkgname=qubes-vm-core
+pkgname=(qubes-vm-core qubes-vm-networking qubes-vm-keyring)
```
marmarek: How would it work on update? Wouldn't ppl end up without qubes-vm-networking -> broken network?

ptitdoc:
I pushed the branch before pushing the signed tag. As a result code signing appears to fail...

archlinux/PKGBUILD
marmarek: Ping @ptitdoc 
Answer for this question apply to other waiting PRs too...

ptitdoc: You are right right about the package qubes-vm-networking not being installed by default.

The qubes-vm-networking package include all the tools/dependencies for using archlinux as a netvm (firewall ...), however, basic networking (setup-ip...) is present in the qubes-vm-core agent.

For QubesR3 however, qubes-vm-networking is only a stub package and only reference the required dependencies for using the netvm tools/scripts.

PR #69, opened 2017-11-13
Fix the Archlinux template update proxy to work for HTTPS URLs as well

na--:
I was getting more errors than expected when updating the template and it was because I'd used an HTTPS pacman repository. [This](https://curl.haxx.se/docs/manpage.html#ENVIRONMENT) is the curl documentation on proxy configuration. I tested the change with my template and it worked.


na--:
I think that this does not conflict with @ptitdoc's pending pull requests, but if it does I have no problems rebasing it or making a pull request to his branch. And it also should be suitable for 3.2, though I have no easy way to test that.

PR #71, opened 2017-11-14
Fix Ubuntu template builds

noptys:
4.0 template builds use `<package>.install` files with dh_install.  The
differences between Debian and Ubuntu packages also need to be represented
in these files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -302,6 +302,8 @@ endif
 
 ifeq ($(shell lsb_release -is), Debian)
 	install -m 0644 misc/qubesxdg.py $(DESTDIR)/$(PYTHON2_SITELIB)/
+else ifeq ($(shell lsb_release -is), Ubuntu)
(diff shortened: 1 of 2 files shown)


marmarek:
Oh, @unman already fixed this in #70, but this one is also a good idea, to make this future-proof.

unman:
It wont necessarily make things future proof because in many cases the position changes from one Ubuntu version to the next, e.g both trusty and zesty build and run without what was a Xenial specific change. It may or may not work in future.

noptys:
As it stands, there is duplicate logic between `Makefile` and `Makefile.builder` for determining what files to be included in each package depending on the specific version of Debian or Ubuntu.  I tried an approach of having a common `debian/qubes-core-agent.install` file and appending to it, but it seemed pretty ugly.  Suggestions?

PR #72, opened 2017-11-18
Flavors

fepitre:


marmarek:
See travis failed builds:
 - you forgot to update debian packaging
 - shellcheck complains - for `'$default'` and `$files` add appropriate `shellcheck disable=` comment, but `$file` should be quoted.

fepitre:
If you have time, could you also test the script for DVM cases? I need to dig why I have connection refused when I try to have a DVM in rc2...

qubes-rpc/qvm-actions.sh
```diff
@@ -0,0 +1,54 @@
...
+fi
+
+# File(s)
+files=${*:2}
+
+# copy and move handle a list of files where other actions don't
+case $1 in
+    copy)
+        #shellcheck disable=SC2016
+        qvm-copy-to-vm '$default' "$files" | zenity --notification --text="Copying files..." --timeout 3
```
marmarek: `$files` should not be quoted, otherwise you'll give all the files as a single argument.
Better approach to handle file names with spaces is to use `"$@"` and `shift`, like this:
```
action="$1"
shift
(...)
qvm-copy-to-vm '$default' "$@"
```

misc/uca_qubes.xml
```diff
@@ -0,0 +1,72 @@
...
+	<image-files/>
+	<other-files/>
+	<text-files/>
+	<video-files/>
+</action>
+<action>
+	<icon>gtk-convert</icon>
+	<name>Convert in DisposableVM</name>
+	<unique-id>1507455488971315-6</unique-id>
+	<command>/usr/lib/qubes/qvm-actions.sh p %F</command>
```
marmarek: Should be `pdf`.

qubes-rpc/qvm-actions.sh
```diff
@@ -0,0 +1,57 @@
...
+    copy)
+        #shellcheck disable=SC2016
+        qvm-copy-to-vm '$default' "$@" | zenity --notification --text="Copying files..." --timeout 3
+        ;;
+    move)
+        #shellcheck disable=SC2016
+        qvm-move-to-vm '$default' "$@" | zenity --notification --text="Moving files..." --timeout 3
+        ;;
+    img)
+        for file in $files
```
marmarek: "$@" here should do here too (including proper whitespaces handling).

fepitre: You want me to drop the $files and replace it by $@? I can do it but with $* to take into account the newline IFS. I tried $@ and it does as usual, it splits when there is spaces in filenames.

PR #73, opened 2017-11-19
Add iptables dep to qubes-core-vm RPM spec

rustybird:
For R3.2


rustybird:
Bumping this in case you forgot to push your local `release3.2` branch after approving the PR.

PR #76, opened 2017-12-12
Fix language issues and usability issue

zander:
The generated default scripts should have better English grammar and as the number one usability issue with changing them is that they were not executable, make them so from the start.


init/setup-rw.sh
```diff
@@ -25,33 +25,31 @@ if mountpoint -q /rw ; then
...
 #  ln -s /rw/config/cups /etc/cups
 #  systemctl --no-block restart cups
 EOF
+        chmod 755 /rw/config/rc.local
 
         touch /rw/config/qubes-firewall-user-script
         cat > /rw/config/qubes-firewall-user-script <<EOF
 #!/bin/sh
 
-# This script is called in ProxyVM after firewall every update (configuration
```
marmarek: "firewall" word was important here - this script isn't called on any update, just after firewall change.

zander: Ah, makes sense.

I force-pushed a new version re-adding the word "firewall"

PR #77, opened 2017-12-13
qvm-{copy,move}: fix spurious deprecation message

rustybird:
`qvm-{copy,move}` run `qvm-{copy,move}-to-vm` with `$default` as the VM argument. Don't print the deprecation message in that case.


rustybird:
Force pushed with `\$default` instead of `'$default'`, to placate shellcheck.

PR #80, opened 2017-12-16
xdg dbus

newchain:
I have been using this with a dbus-less Gentoo template since the original
change, and have tested recently on whonix-gw with dbus enabled and running.

### DIFF
--- a/misc/xdg.py
+++ b/misc/xdg.py
@@ -2,18 +2,25 @@
 
 from gi.repository import Gio
 import sys
-import dbus
 
 def launch(desktop, *files):
     launcher = Gio.DesktopAppInfo.new_from_filename(desktop)
-    if hasattr(launcher, 'get_boolean'):
-        activatable = launcher.get_boolean('DBusActivatable')
-        if activatable:
-            bus = dbus.SessionBus()
-            service_id = launcher.get_id()
-            # cut the .desktop suffix
-            service_id = service_id[:-8]
-            bus.start_service_by_name(service_id)
+    try:
+        import dbus
+        if hasattr(launcher, 'get_boolean'):
+            activatable = launcher.get_boolean('DBusActivatable')
+            if activatable:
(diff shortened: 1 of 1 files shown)


misc/xdg.py
```diff
@@ -2,19 +2,25 @@
...
+                service_id = launcher.get_id()
+                # cut the .desktop suffix
+                service_id = service_id[:-8]
+                try:
+                    bus.start_service_by_name(service_id)
+                except dbus.DBusException:
+                    launcher.launch(files, None)
+        launcher.launch(files, None)
+    except:
+        launcher.launch(files, None)
```
marmarek: This is wrong:
 - bare `except` (without specifying exception) is a bad idea in most cases - you may handle this way something totally different that you wanted
 - in worst case, you'll launch the same action thrice

Better use something like:
```python
try:
    import dbus
    # ... try dbus, return from the function if succeed
except ImportError:
    pass
launcher.launch(files, None)
```

marmarek:
> I have been using this with a dbus-less Gentoo template

I guess this overlay: https://github.com/newchain/qubes-overlay, right? How does it work? Maybe you want to add info about it to [the docs](https://www.qubes-os.org/doc/templates/)?

misc/xdg.py
newchain: Is this better?

newchain:
> I guess this overlay: https://github.com/newchain/qubes-overlay, right? How does it work? Maybe you want to add info about it to the docs?
It probably will not work at the moment because Gentoo's version of xen-utils is 4.9 and so your patches may not apply. But using it is simple - install Gentoo, put the overlay somewhere and point to it with /etc/portage/repos.conf.

marmarek:
Note for the future: please open PR on master branch, then note you want it also backported to release3.2 (or elsewhere). For now I've cherry picked it to master.

PR #81, opened 2017-12-26
Flavor xfce r3.2

fepitre:
- Please Travis do the builds :)
- Please check also the debian packaging.


marmarek:
Travis said: debian packaging is broken

fepitre:
What is the difference between this branch and the master branch? I'm not so familiar with debian packaging and I'am not sure to understand what's wrong. I inserted the same part than the master branch. But there is no others packages with .install .postinst ... like for nautilus or others. I seems that what is in qubes-core-agent-thunar.install is skipped.

marmarek:
The difference is that in release3.2 branch there were no other
subpackages, so all files are installed directly to qubes-core-agent
package (export DESTDIR=$(shell pwd)/debian/qubes-core-agent) instead of
temporary place (export DESTDIR=$(shell pwd)/debian/tmp) and use install
files to select individual files. See debian/rules. So, you need to:

1. Change debian/rules to use debian/tmp
2. Add debian/qubes-core-agent.install with a list of files there (you
can download package from
https://deb.qubes-os.org/r3.2/vm/pool/main/q/qubes-core-agent/qubes-core-agent_3.2.22-1+deb8u1_amd64.deb
and use dpkg-deb to list files).

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

fepitre:
From the list of files in the deb package, i removed:

usr/share/glib-2.0/schemas/10_qubes-core-agent.gschema.override
usr/lib/qubes/bind-dirs.sh
usr/share/doc/qubes-core-agent/changelog.Debian.gz
usr/share/doc/qubes-core-agent/copyright

and then put it in the qubes-core-agent.install.

rpm_spec/core-vm.spec
```diff
@@ -198,6 +198,13 @@ DNF plugin for Qubes specific post-installation actions:
  * refresh applications shortcut list
 %endif
 
+%package thunar
```
marmarek: Base package name in r3.2 is `qubes-core-vm`, so this will result in `qubes-core-vm-thunar` package. This is not what builder-fedora expects. That was renamed to qubes-core-agent in r4.0 (to be consistent with debian and repository name).
So, here better use `%package -n qubes-core-agent-thunar` (and similar for other % macros) to have the same package name everywhere.

fepitre:
Arg. Forgot "-n". I do the correction.

PR #84, opened 2018-01-14
qubes.VMShell+WaitForSession (VM side)

rustybird:


rustybird:
Note that this was force pushed once. I always, always forget to add new files to the Debian `.install` list.

PR #88, opened 2018-02-06
Stop Debian templates from forwarding by default.

unman:
Debian templates currently have ip_forward set to 1.
This PR changes that default.

Closes QubesOS/qubes-issues#3453


marmarek:
Could you open PR against master?

PR #89, opened 2018-02-06
Stop Debian templates from forwarding by default.

unman:
Debian templates currently have ip_forward set to 1.
This PR changes that default.

Closes QubesOS/qubes-issues#3453


network/80-qubes.conf
```diff
@@ -1,2 +0,0 @@
-net.ipv4.ip_forward=1
-net.ipv6.conf.all.drop_unsolicited_na=1
```
marmarek: This other line should remain...

unman: It should indeed

PR #91, opened 2018-02-08
bind-dirs.sh: don't fail on empty configuration directory

rustybird:
`nullglob` must be active before a glob is used on a potentially empty directory. Call `shopt` immediately after the shebang line.

Fixes QubesOS/qubes-issues#3552


lvh:
LGTM (I'm the reporter of the referenced bug).

PR #95, opened 2018-02-13
Add qubes-firewall.d feature

tasket:
Enables easier configuration of firewall for vpn tunnels, etc.

### DIFF
--- a/qubesagent/firewall.py
+++ b/qubesagent/firewall.py
@@ -64,14 +64,16 @@ class FirewallWorker(object):
 
     def run_firewall_dir(self):
         '''Run scripts dir contents, before user script'''
-        script_dir_path = '/rw/config/qubes-firewall.d'
-        if not os.path.isdir(script_dir_path):
-            return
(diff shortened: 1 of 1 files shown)


tasket:
https://github.com/QubesOS/qubes-issues/issues/3503

qubesagent/firewall.py
```diff
@@ -62,6 +62,17 @@ def apply_rules(self, source_addr, rules):
         '''Apply rules in given source address'''
         raise NotImplementedError
 
+    def run_firewall_dir(self):
+        '''Run scripts dir contents, before user script'''
+        script_dir_path = '/rw/config/qubes-firewall.d'
```
marmarek: Does it make sense to look _also_ for `/etc/qubes/qubes-firewall.d`? For plugging scripts shipped by some package in template. OTOH, such packages can ship own systemd units and set appropriate ordering (like `Before=network.target` and `After=qubes-firewall.service`)

tasket: Sure!

PR #96, opened 2018-02-14
Really enable qubes-sync-time.timer

rustybird:
/cc @adrelanos, in case you want to disable the timer in Whonix with a drop-in.

### DIFF
--- a/vm-systemd/qubes-sync-time.timer
+++ b/vm-systemd/qubes-sync-time.timer
@@ -6,4 +6,5 @@ ConditionPathExists=!/var/run/qubes-service/clocksync
 OnBootSec=10s
 OnUnitActiveSec=6h
 
-
+[Install]
+WantedBy=multi-user.target


adrelanos:
Rusty Bird:
> /cc @adrelanos, in case you want to disable the timer in Whonix with a drop-in.
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/QubesOS/qubes-core-agent-linux/pull/96

Thanks!

I guess since it has

ConditionPathExists=!/var/run/qubes-service/clocksync

which is not enabled by default for Whonix VMs, on the Whonix side is
nothing to do?

marmarek:
Note that condition is negated - the timer _is_ started when clocksync service is disabled.
But qrexec policy deny `qubes.GetDate` service access for Whonix VMs (those with `anon-vm` tag), so it should be ok.

rustybird:
> But qrexec policy deny `qubes.GetDate` service access for Whonix VMs (those with `anon-vm` tag)

Does it really? I only have this kind of exception in my `qubes.UpdateProxy` policy, although maybe I removed it and forgot.

rustybird:
Sorry, I wanted to cancel the comment, not close the PR. Looks like @marmarek is right.

PR #99, opened 2018-02-23
Fix qubes-dom0-update regression by restoring dom0-updates dir

ij1:
This dir was present in 3.2.22 but then got removed from the package. Missing that dir causes a problem for qubes-dom0-update that expects to find that dir existing.

Fixes: QubesOS/qubes-issues#3620


ij1:
Blah, the ownership is not correct using this approach it seems.

ij1:
Ah, I was wrong with that thought. Its then post-install that seems to adjust the owner as long as the dir exists in the package (so it actually works after all).

PR #110, opened 2018-04-07
Fix dom0 salt not being able to install packages when UpdateVM only has yum installed

vladimir-lu:
See QubesOS/qubes-issues#2223

### DIFF
--- a/misc/qubes-download-dom0-updates.sh
+++ b/misc/qubes-download-dom0-updates.sh
@@ -53,6 +53,9 @@ fi
 YUM="yum"
 if type dnf >/dev/null 2>&1; then
     YUM="dnf --best --allowerasing --noplugins"
+else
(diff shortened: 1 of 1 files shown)


misc/qubes-download-dom0-updates.sh
```diff
@@ -53,6 +53,9 @@ fi
 YUM="yum"
 if type dnf >/dev/null 2>&1; then
     YUM="dnf --best --allowerasing --noplugins"
+else
+    # salt in dom0 thinks it's using dnf but we only have yum so need to remove extra options
+    OPTS="${OPTS/--best --allowerasing/}
```
marmarek: Unclosed "

vladimir-lu: Argh - fixed

PR #115, opened 2018-05-01
Handle multiple 'dconf' packaging who include 'user' file.

fepitre:
For RPM distros, only CentOS and Fedora >=28 include it.


marmarek:
Since this is more and more common that distribution already provide this file, I'm going to drop it entirely here and crate dynamically in %post if not present. This will also solve https://github.com/QubesOS/qubes-issues/issues/3834

PR #121, opened 2018-06-12
Move code to core-qrexec

woju:
This is to be merged after accepting core-qrexec.


marmarek:
Done in https://github.com/QubesOS/qubes-core-agent-linux/commit/20285bc6c272f136edc301e3e024b959efb7f0af

PR #122, opened 2018-06-24
grub: add noresume to kernel cmdline

awokd:
Under R4.0, when Debian HVMs are created from the debian-9 template, they hang on boot for 30 seconds without this option.

https://github.com/QubesOS/qubes-issues/issues/3178


awokd:
Closing to resubmit correctly.

PR #125, opened 2018-07-11
bind mount /usr/local

reynir:
This is a twin PR. Please also see the PR in qubes-linux-template-builder https://github.com/QubesOS/qubes-linux-template-builder/pull/12.

This PR along with its twin changes `/usr/local` from a symlink to a bind mount similar to how `/home` is mounted.

I stumbled upon this because the sandbox in the opam package manager does not mount `/rw/usrlocal`. This meant that `/usr/local` was a dead symlink and that binaries in `/usr/local/bin` was not in PATH inside the sandbox. See https://github.com/ocaml/opam/pull/3446.

### DIFF
--- a/misc/fstab
+++ b/misc/fstab
@@ -4,6 +4,7 @@
 /dev/mapper/dmroot /                       ext4 defaults,noatime        1 1
 /dev/xvdb		/rw			auto	noauto,defaults,discard	1 2
 /rw/home        /home       none    noauto,bind,defaults 0 0
+/rw/usrlocal        /usr/local       none    noauto,bind,defaults 0 0
 /dev/xvdc1      swap                    swap    defaults        0 0
 tmpfs                   /dev/shm                tmpfs   defaults,size=1G        0 0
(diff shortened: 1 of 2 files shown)


init/setup-rw.sh
```diff
@@ -72,6 +72,13 @@ if [ -L /home ]; then
     mkdir /home
 fi
 
+# Old Qubes versions had symlink /usr/local -> /rw/usrlocal; now we use mount --bind
+if [ -L /usr/local ]; then
+    rm /usr/local
+    mkdir /usr/local
+fi
+
+
```
marmarek: This should be at packaging level (rpm spec, debian postinst, archilinux install scripts), not at every VM start.

reynir: Ok. I went with how the `/home` upgrade seemingly was handled. I'm not sure how you mean. Should I remove this commit? Thanks

marmarek: Yes, please.

And when you pointed /home handling there I think I should move it to package scripts too. Or remove completely (we've changed /home to a bind mount in 2012...).

PR #127, opened 2018-07-25
Tweak graphical file copy/move

rustybird:


rustybird:
@marmarek "Unable to verify (no valid key found)" - not sure what's up, the commits are signed as usual. I assumed there was some keyserver flakiness and repushed, but it happened again.

PR #128, opened 2018-07-26
Install qubes control files for services in Ubuntu templates

unman:
This effectively reverts the commits in https://github.com/QubesOS/qubes-core-agent-linux/commit/80b5c942069710f75aa7faea141e31c1cba60b33 and https://github.com/QubesOS/qubes-core-agent-linux/commit/54867b6eabed3a6dca0e932ff77f67aedfe43e03 which now prevent Xenial (and later) templates from being built.
There's no reason why those control files should not be installed.

Closes QubesOS/qubes-issues/issues/3871


unman:
@marmarek I think there was a timing error between my uploading the new key and the policy check.
The commit is showing as "verified" now.

PR #129, opened 2018-08-07
Handle CentOS vm repo

fepitre:


Makefile
```diff
@@ -161,7 +161,8 @@ install-sysvinit: install-init
 	install network/qubes-iptables $(DESTDIR)/etc/init.d/
 
 install-rh: install-systemd install-systemd-dropins install-sysvinit
-	install -D -m 0644 misc/qubes-r4.repo $(DESTDIR)/etc/yum.repos.d/qubes-r4.repo
+	install -D -m 0644 misc/qubes-r4.repo.in $(DESTDIR)/etc/yum.repos.d/qubes-r4.repo
+	sed -i "s/@DIST@/$(shell echo $(DIST) | sed 's/[[:digit:]]//g')/g" $(DESTDIR)/etc/yum.repos.d/qubes-r4.repo
```
marmarek: what about `${DIST%%[0-9]*}` instead of `$(shell ...)`?

fepitre: Oh I didn't know such syntax in Makefile, in Bash I would have used ${DIST//[[:digit:]]}. So I tried but got empty result. Will check.

marmarek: Yes, I meant shell syntax for that. You may need to use `$$` instead of `$`.

marmarek:
Ok, so leave as is (but squash into the first commit).

fepitre:
Done.

PR #130, opened 2018-08-08
archlinux: ensure /usr/local is present in fstab

ptitdoc:
Complete fix of the issue https://github.com/QubesOS/qubes-issues/issues/1150 for archlinux


ptitdoc:
I pushed a signed tag after the travis check.

PR #132, opened 2018-08-12
Fix Ubuntu build in 3.2

unman:


unman:
@marmarek I'm less bothered by the build errors, (looks like network issue), than the "no valid key".
I dont know where the check is, but either an old copy is cached or it's pulling from a keyserver that isn't synced.
I've checked on a few, and always get the updated key.

PR #137, opened 2018-10-01
Voice informational messages in bind-dirs.sh

lvh:
Previously, bind-dirs.sh had a bunch of `true comment goes here` style debug messages (no-ops). Presumably this was done because these messages are intended as debug messages and would only be displayed when calling bind-dirs.sh with xtrace enabled. However, this includes some fatal errors, which are necessary to debug why bind-dirs.sh is ostensibly not working. For example, I tried to mount /var/lib/docker, didn't realize it did not exist (as an empty directory) in my base template, and there was no journalctl output at all. After this change, journalctl will contain the (very helpful) error message.

### DIFF
--- a/vm-systemd/bind-dirs.sh
+++ b/vm-systemd/bind-dirs.sh
@@ -31,7 +31,7 @@ source /usr/lib/qubes/init/functions
 
 prerequisite() {
    if ! is_rwonly_persistent ; then
-      true "No TemplateBasedVM detected. Exiting."
+      echo "No TemplateBasedVM detected. Exiting."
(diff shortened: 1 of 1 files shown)


marmarek:
Generally better to output diagnostics to stderr (`>&2`). But if the script doesn't produce "actual" output, that's fine too.

lvh:
FWIW I don't have merge access andI'm not sure who does :-)

marmarek:
Don't worry, I'll merge it. It's just that I have some in-progress development in this repository and don't want to mix those two.

PR #141, opened 2018-10-24
Allow per-VM protected file list

Rudd-O:
Hopefully this can be pushed as an update for Qubes 3.2 as well?

### DIFF
--- a/init/functions
+++ b/init/functions
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Location of files which contains list of protected files
-PROTECTED_FILE_LIST='/etc/qubes/protected-files.d'
+PROTECTED_FILE_LIST='/etc/qubes/protected-files.d /rw/qubes/protected-files.d'
 
 qsvc() {
     # Returns whether a service is enabled.
@@ -100,7 +100,7 @@ reload_random_seed() {
 }
 
 is_protected_file() {
-    grep -Fxrq --exclude='*.rpmsave' --exclude='*~' --exclude='*.rpmnew' --exclude='*.rpmold' -- "${1}" "$PROTECTED_FILE_LIST" 2>/dev/null
+    grep -Fxrq --exclude='*.rpmsave' --exclude='*~' --exclude='*.rpmnew' --exclude='*.rpmold' -- "${1}" $PROTECTED_FILE_LIST 2>/dev/null
 }
 
 umount_retry() {
--- a/init/functions
+++ b/init/functions
@@ -101,7 +101,7 @@ reload_random_seed() {
 }
 
 is_protected_file() {
-    local ret=0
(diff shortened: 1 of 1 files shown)


init/functions
```diff
@@ -100,7 +101,16 @@ reload_random_seed() {
...
 
 is_protected_file() {
-    grep -Fxrq --exclude='*.rpmsave' --exclude='*~' --exclude='*.rpmnew' --exclude='*.rpmold' -- "${1}" "$PROTECTED_FILE_LIST" 2>/dev/null
+    local ret=0
+    local pfilelist
+    for pfilelist in "$PROTECTED_FILE_LIST" "$PER_VM_PROTECTED_FILE_LIST" ; do
+        if test -d "$pfilelist" ; then
+            # If this succeeds, we return immediately to the caller.
+            # If not, we let the loop continue.
+            grep -Fxrq --exclude='*.rpmsave' --exclude='*~' --exclude='*.rpmnew' --exclude='*.rpmold' -- "${1}" "$pfilelist" 2>/dev/null && return 0 || ret="$?"
```
Rudd-O: We also make the code compatible with `-e` setting in Bash.

init/functions
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 
 # Location of files which contains list of protected files
-PROTECTED_FILE_LIST='/etc/qubes/protected-files.d /rw/qubes/protected-files.d'
+PROTECTED_FILE_LIST='/etc/qubes/protected-files.d'
+PER_VM_PROTECTED_FILE_LIST='/rw/qubes/protected-files.d'
```
marmarek: We have /rw/config for configuration on /rw

Rudd-O: OK, the patch can be altered to use that.  Done.

Rudd-O:
I see uses of `is_protected_file` in `qubes-early-vm-config.sh` whose systemd unit `vm-systemd/qubes-early-vm-config.service` explicitly says `After=local-fs.target` which I believe means after `/rw` has been mounted.

I see uses of it in `setup-ip` which happens way after during boot.

I see uses of it in the initscript deployed by `qubes-core-agent.postinst`, and initscripts run pretty much after all file systems have been mounted.

Am I missing something important, @marmarek ?

init/functions
```diff
@@ -100,7 +101,16 @@ reload_random_seed() {
 }
 
 is_protected_file() {
-    grep -Fxrq --exclude='*.rpmsave' --exclude='*~' --exclude='*.rpmnew' --exclude='*.rpmold' -- "${1}" "$PROTECTED_FILE_LIST" 2>/dev/null
+    local ret=0
```
marmarek: One issue with this: if none of config directories exists, all files will be marked as "protected".

PR #143, opened 2018-11-07
skip inside chroot

adrelanos:
https://github.com/QubesOS/qubes-issues/issues/4469

Untested.

### DIFF
--- a/qubes-rpc/qubes.PostInstall
+++ b/qubes-rpc/qubes.PostInstall
@@ -21,6 +21,10 @@
 #
 #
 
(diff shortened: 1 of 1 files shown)


qubes-rpc/qubes.PostInstall
```diff
@@ -21,6 +21,10 @@
 #
 #
 
+if ischroot --default-false ; then
```
marmarek: Theoretically looks good, but `ischroot` is debian only :(

PR #145, opened 2018-11-29
Rephrase comment

strugee:
Originally came here just to s/dis/disk/, but rephrased while I was at it. I think the new one reads more clearly but happy to amend to just fix the typo.


strugee:
CI failures is spurious and unrelated.

PR #146, opened 2018-12-01
FIx rootfs resize script

strugee:
It was just being compared the wrong way; the rest of these are cleanups. At least, assuming I understand the script right - `non_rootfs_data` is just supposed to be a small margin for error right? I.e. that's what's meant by "(almost) the whole disk"?

It took me a while to get that so I renamed the variable to make it clearer although I'm not sure the new name is _all_ that much better... happy to either drop that commit or amend it to something else.

Fixes QubesOS/qubes-issues#4553 (and improves VM boot performance a _lot_ in the process; it shaves several entire seconds off boot times which especially adds up when all autostart VMs are being brought up at dom0 boot)

### DIFF
--- a/init/resize-rootfs-if-needed.sh
+++ b/init/resize-rootfs-if-needed.sh
@@ -14,7 +14,9 @@ fi
 sysfs_xvda="/sys/class/block/xvda"
 
 # if root filesystem use already (almost) the whole dis
-size_margin=$(( 250 * 1024 * 2 ))
+# 203M for BIOS and /boot data, 222 for ext4 filesystem overhead
+# See QubesOS/qubes-core-agent-linux#146 for more details
+size_margin=$(((222 + 203) * 2 * 1024))
 rootfs_size=$(df --block-size=512 --output=size / | tail -n 1 | tr -d ' ')
 if [ $(cat $sysfs_xvda/size) -lt \
        $(( size_margin + rootfs_size )) ]; then
--- a/init/resize-rootfs-if-needed.sh
+++ b/init/resize-rootfs-if-needed.sh
@@ -15,7 +15,7 @@ sysfs_xvda="/sys/class/block/xvda"
 
 # if root filesystem use already (almost) the whole dis
 non_rootfs_data=$(( 250 * 1024 * 2 ))
-rootfs_size=$(df --block-size=512 --output=size / | tail -n 1)
+rootfs_size=$(df --block-size=512 --output=size / | tail -n 1 | tr -d ' ')
 if [ $(cat $sysfs_xvda/size) -gt \
        $(( non_rootfs_data + rootfs_size )) ]; then
    echo "root filesystem already at $rootfs_size blocks" >&2
--- a/init/resize-rootfs-if-needed.sh
+++ b/init/resize-rootfs-if-needed.sh
@@ -16,7 +16,7 @@ sysfs_xvda="/sys/class/block/xvda"
 # if root filesystem use already (almost) the whole dis
 non_rootfs_data=$(( 250 * 1024 * 2 ))
 rootfs_size=$(df --block-size=512 --output=size / | tail -n 1)
-if [ "$(cat "$sysfs_xvda/size")" -lt \
+if [ $(cat $sysfs_xvda/size) -lt \
        $(( non_rootfs_data + rootfs_size )) ]; then
    echo "root filesystem already at $rootfs_size blocks" >&2
    exit 0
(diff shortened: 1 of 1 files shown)


strugee:
Didn't realize I hadn't pushed this particular key to the keyserver. I just uploaded it; if you kick the codesigning bot in a bit (once the key has propagated to whatever keyserver the bot uses) it should turn green.

init/resize-rootfs-if-needed.sh
```diff
@@ -14,12 +14,10 @@ fi
...
-non_rootfs_data=$(( 250 * 1024 * 2 ))
-rootfs_size=$(df --output=size / | tail -n 1)
-# convert to 512-byte blocks
-rootfs_size=$(( rootfs_size * 2 ))
-if [ "$(cat "$sysfs_xvda/size")" -lt \
-       $(( non_rootfs_data + rootfs_size )) ]; then
+size_margin=$(( 250 * 1024 * 2 ))
+rootfs_size=$(df --block-size=512 --output=size / | tail -n 1 | tr -d ' ')
+if [ $(cat $sysfs_xvda/size) -gt \
+       $(( size_margin + rootfs_size )) ]; then
```
marmarek: This change looks wrong.
We have two sizes here:
 - size of actual disk (read from sysfs)
 - size of filesystem on it (read with df)

The script should resize root filesystem if the disk beneath it was extended. In other words - if the filesystem does not span the whole disk. The margin here is to take into account various metadata (partition table etc) and boot partitions (/boot and bios boot), which by default are 200M + 2M.

The condition here is about _skipping_ the resize, so it should be the other way around - skip if disk is _not_ larger than filesystem. So it definitely should stay `-lt` (or `-le`).

I think the actual problem is either with `df` size reporting (unlikely), or the size margin being too small (something not taken into account?).

strugee:
> I've rerun the signature check, but it still can't find the key.

Hmm, I submitted it to `keyserver.ubuntu.com` - I thought that synchronized with the global network but maybe not? I pushed it to `keys.gnupg.net` this time. Hopefully that'll work.

init/resize-rootfs-if-needed.sh
strugee: Oh wow, yep. I totally misread. I'll poke at it some more.

strugee: OK so, I've dug into this a little more. Here's what I've found:

* ~203 MB is definitely the right size margin (confirmed by `echo '(' $(cat /sys/block/xvda/size) - $(cat /sys/block/xvda/xvda3/size) ')' / 2 / 1024 | bc`)
* However, setting `size_margin` to 204M still runs `resize2fs`.
* It seems like a `size_margin` of 203+221 triggers resizing, whereas 203+222 doesn't.

The question is where that 222M of data is coming from. I considered a lot of things (maybe there's a 1000/1024 mismatch somewhere, or maybe it was alignment?) but in the end I think this is probably the filesystem overhead. That is, the space on disk occupied by filesystem datastructures that isn't directly usable to store data.

I checked using `sudo dumpe2fs /dev/xvda3`:

* The journal size is reported as 64M and presumably isn't included in usable free space as reported by `df`; that leaves 158M of data unaccounted for.
* Inode size is reported as 256 (bytes presumably?); 643376 is reported as "inode count". AFAICT that means the number of inodes allocated in the inode table, so 256 * 643376 / 1024^2 ~= 157.074M. Call that 157M, leaving ~1M of data unaccounted for.
* Reserved GDT blocks is reported as 1024; block size is reported as 4096, which means that reserved GDT blocks (there for future filesystem expansion) take up 4M of space. That would account for _more_ than the remaining ~1M of data, not even counting the remaining fixed-width data.

References: `mke2fs(8)`, `dumpe2fs(8)`, https://ext4.wiki.kernel.org/index.php/Ext4_Disk_Layout

Do note though that I'm by _no_ means an ext4 expert so it's very possible I got something wrong. I have, however, pushed a commit using this new number. I'll test it shortly.

marmarek:
I have no idea why signature check still fails - when I do it manually it looks fine.

strugee:
> I have no idea why signature check still fails - when I do it manually it looks fine.

`keys.gnupg.net` is a round-robin kinda deal, maybe the codesigning bot is resolving a different server than you are? If they haven't all synchronized yet? Otherwise I have no idea.

If it still doesn't work after a while are you able to merge anyway?

marmarek:
> If it still doesn't work after a while are you able to merge anyway?

Yes, I can ignore it if I choose to ;)

strugee:
> Yes, I can ignore it if I choose to ;)

Heh, I meant more in terms of policy rather than technical ability :P

Regardless, super stoked to see this go in! Thanks for the very helpful review.

PR #147, opened 2018-12-07
Add XTerm as a dependency

strugee:
Qubes Manager's update button fails in strange ways without it.


strugee:
I'm having a hard time reading the Travis log so I can't tell why it actually failed...

strugee:
CI is passing now; @marmarek I assume you restarted it?

PR #150, opened 2019-01-15
Switch to HTTPS

lunarthegrey:
https://github.com/QubesOS/qubes-issues/issues/4415

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -26,6 +26,7 @@ Vcs-Git: https://github.com/QubesOS/qubes-core-agent-linux
 Package: qubes-core-agent
 Architecture: any
 Depends:
+    apt-transport-https,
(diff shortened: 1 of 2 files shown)


lunarthegrey:
> For this to ship as an update, the package needs to explicitly depends on `apt-transport-https`, otherwise it may break some existing installations.

Ah, my mistake. Is it hard to have apt-transport-https installed with a package upgrade of this? I do not understand Qubes internals very well, so pardon my ignorance. I would be happy to do any file editing grunt-work :)

marmarek:
It's not hard, but you need to add it to debian/control.

lunarthegrey:
@marmarek I've added the dependency to debian/control

PR #152, opened 2019-02-07
Use dumpe2fs for filesystem size calculations

strugee:
The previous approach from #146 used `df` to get usable space and then added a fixed size to that number in order to account for filesystem overhead. However, at some point that stopped working for me. It appears that ext4 filesystem overhead can vary over time or because of other factors. (Certainly now that I think about it the old code would only work well for people with the exact same filesystem size as me.)

So the new approach is to just completely ignore what `df` tells us and instead go directly to the source: the filesystem's internal notion of exactly how much space it takes up. We use `dumpe2fs` to retrieve this information and calculate the on-disk size dynamically from that. Then we add the space that boot data takes up (unchanged), and we add 5MB padding because when I tested this it didn't quite add up otherwise. https://unix.stackexchange.com/a/13551/29146 suggests that this unaccounted-for data may be e.g. additional copies of the superblock.

Initial testing looks good for this but I want to do some additional testing before I call it good to go. Also both invocations of `dumpe2fs` print a header on stderr which is pretty annoying. I thought about trying to filter it out by redirecting it to stdout (and thus into `grep`) but then I would've been filtering out legitimate error messages, if there were any. I could do some shell wrangling to pipe stderr into a separate `grep -v` process to get rid of the header, but at that point it just didn't seem worth it.

### DIFF
--- a/init/resize-rootfs-if-needed.sh
+++ b/init/resize-rootfs-if-needed.sh
@@ -14,12 +14,20 @@ fi
 sysfs_xvda="/sys/class/block/xvda"
 
 # if root filesystem is already using (almost) the whole disk
-# 203M for BIOS and /boot data, 222 for ext4 filesystem overhead
-# See QubesOS/qubes-core-agent-linux#146 for more details
-size_margin=$(((222 + 203) * 2 * 1024))
-rootfs_size=$(df --block-size=512 --output=size / | tail -n 1 | tr -d ' ')
+# 203M for BIOS and /boot data
+boot_data_size=$((203 * 2 * 1024))
+# rootfs size is calculated on-the-fly. `df` doesn't work because it doesn't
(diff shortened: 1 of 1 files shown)


strugee:
My key expired yesterday; I pushed an update to the SKS keyserver pool but it definitely hasn't propagated everywhere yet. Kick the bot in a little while and it'll work, or you can try `keyserver.ubuntu.com` which I also pushed to (will work _now_ because it's an individual server, not a pool).

strugee:
Initial testing shows that if the qube size is increased by 20 MB there is a kernel panic because the root filesystem can't be mounted. This does appear to be a regression; however I'm still investigating. 100MB increase works fine.

strugee:
Update: the bug was not a regression caused by this PR. I will file an issue. (I can't reproduce in a Fedora 28 VM.)

PR #155, opened 2019-04-14
Handle tinyproxy path changed since Fedora 30+

fepitre:
QubesOS/qubes-issues#4973


marmarek:
As @unman noted in #4973, this also affects Debian buster. In this case IMO we need some more generic solution, not limited to rpm. Maybe runtime discovery wasn't such a bad idea after all...

fepitre:
@marmarek @unman, here is a new version: https://github.com/fepitre/qubes-core-agent-linux/commit/0fd8da62b6499945531707f59392cdb5d8ef0b8e maybe some packaging fixes remain

fepitre:
I close this PR and will reopen on another branch.

PR #158, opened 2019-05-05
Showterminal

fepitre:


debian/qubes-core-agent.install
```diff
@@ -8,6 +8,7 @@ etc/default/grub.d/30-qubes.cfg
 etc/fstab
 etc/needrestart/conf.d/50_qubes.conf
 etc/profile.d/qt_x11_no_mitshm.sh
+etc/qubes-rpc/admin.vm.TerminalDispVM
```
marmarek: See https://github.com/QubesOS/qubes-core-admin/pull/259#discussion_r281393567 comment about the name.

qubes-rpc/admin.vm.TerminalDispVM
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash --
+
+set -eu
+sock_name="$(mktemp -u XXXXXX)"
```
marmarek: Better use random directory in `/tmp` and socket within it. `/var/run/` is for runtime data, not temporary files/sockets.

qubes-rpc/admin.vm.TerminalDispVM
```diff
@@ -0,0 +1,17 @@
...
+if [[ "$sock_name" =~ [^a-zA-Z0-9] ]]; then
```
marmarek: What are you checking here? if `mktemp` worked? That should be handled by `set -e` already.

qubes-rpc/admin.vm.TerminalDispVM
```diff
@@ -0,0 +1,17 @@
...
+  echo "Invalid internal sock name provided."
+  exit 1
+fi
+sock="/var/run/qubes/$sock_name.terminal.sock"
+
+xterm -geometry 80x24 -e /bin/sh -c "
+until [ -S $sock ]; do sleep 1; done || true
```
marmarek: I'd put smaller sleep here. One may want to catch early boot...
Also, it may be good to add echo with a message how to exit from this.
But on the other hand, in this case closing the window should be enough.

debian/qubes-core-agent.install
```diff
@@ -100,6 +101,7 @@ usr/bin/qvm-move-to-vm
 usr/bin/qvm-open-in-dvm
 usr/bin/qvm-open-in-vm
 usr/bin/qvm-run-vm
+usr/bin/qvm-terminal
```
marmarek: I wonder if `qvm-console` wouldn't be a better name. There is `xl console`, `virsh console`, etc.
Then, same for `admin.vm.Terminal` -> `admin.vm.Console`.

qubes-rpc/qubes.ShowInTerminal
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash --
+
+set -eu
+tmpdir="$(mktemp -d)"
+sock="${tmpdir//\"/\"\\\"\"/}/terminal.sock"
```
marmarek: This looks scary. And also is wrong - if path contains `"`, it will fail in the first socat call this way. `"` should not appear here anyway. IMO better only check and exit with an error if that happens, instead of trying to handle this situation fully.

PR #159, opened 2019-05-16
Handle non-default 'eth0' Qubes managed interface

fepitre:


init/functions
```diff
@@ -128,6 +128,25 @@ umount_retry() {
...
+    iface="$(ip -o link | grep -i "$mac" | awk '{print $2}' | cut -d ':' -f1)"
+    echo "$iface"
+}
+
+get_qubes_managed_iface() {
+    local mac
+    local qubes_iface
+    mac="$(qubesdb-read /qubes-mac)"
+    qubes_iface="$(get_iface_from_mac "$mac")"
+    if [ "x$qubes_iface" != "x" ]; then
```
marmarek: Please test if that works. For me it looks like it doesn't: missing `/qubes-mac` results in empty `$mac`, which in turns will result in `grep -i ""` in `get_iface_from_mac`, so it will return all the interfaces...

marmarek:
One more thing: you had an issue when `setup-ip` was called by udev rules, for all the vif interfaces. To solve this problem, you may modify `setup-ip` script to check if `/qubes-mac` is present and if it is, check if interface to configure has this MAC.

init/functions
fepitre: I can add one more check on mac variable but why would this happen?

marmarek: As said above - I think `get_iface_from_mac` called with empty mac will return list of all interfaces, not empty result.

fepitre: Sure, I was just asking if there were a reason/possibility that /qubes-mac path would be empty.

network/setup-ip
```diff
@@ -6,20 +6,24 @@
...
-    /sbin/ethtool -K "$INTERFACE" sg off
-    /sbin/ethtool -K "$INTERFACE" tx off
-    # If NetworkManager is enabled, let it configure the network
-    if qsvc network-manager ; then
-        nm_config=/etc/NetworkManager/system-connections/qubes-uplink-$INTERFACE
-        cat > "$nm_config" <<__EOF__
+mac="$(/usr/bin/qubesdb-read /qubes-mac 2> /dev/null)"
+current_mac="$(get_mac_from_iface "$INTERFACE")"
+
+if [ "$mac" = "$current_mac" ]; then
```
marmarek: This breaks the network, if `/qubes-mac` is missing.
Please test how it works without dom0 change in place too.

marmarek:
Now this looks better, and simple manual test says it works. But I'll run full automated test suite before merging it.

marmarek:
Network tests succeeded: https://openqa.qubes-os.org/tests/2363

PR #160, opened 2019-05-30
Handle network hooks located in /rw/config/network-hooks.d

fepitre:
Example:
/rw/config/network-hooks.d/test.sh
\#!/bin/bash

command="$1"
vif="$2"
ip="$3"

if [ "$ip" == '10.137.0.100' ]; then
    case "$command" in
        online)
            ip route add 192.168.0.100 via 10.137.0.100
            ;;
        offline)
            ip route del 192.168.0.100
            ;;
    esac
fi

### DIFF
--- a/network/vif-route-qubes
+++ b/network/vif-route-qubes
@@ -26,6 +26,26 @@ dir=$(dirname "$0")
 
 #main_ip=$(dom0_ip)
 
+# Network Hooks for triggering supplementary actions on AppVM connect
+network_hooks() {
+    local command="$1"
+    local vif="$2"
+    local addr="$3"
+
+    local vif_type
+    vif_type="$(xenstore-read "${XENBUS_PATH}/type")"
+
+    if [ -d /rw/config/network-hooks.d ]; then
+        for hook in /rw/config/network-hooks.d/*
+        do
+            if [ -x "$hook" ]; then
+                log debug "Executing network-hook $(basename "$hook")..."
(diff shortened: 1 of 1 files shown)


fepitre:
I also cleanup the tabs by replacing them with spaces.

marmarek:
As for metric, maybe just export that variable. I don't think it should be part of an API we'll need to keep stable.

network/vif-route-qubes
```diff
@@ -95,19 +95,29 @@ domid=${domid/.*/}
...
         echo -e "*raw\\n$iptables_cmd -i ${vif} ! -s ${addr} -j DROP\\nCOMMIT" | \
             ${cmdprefix} $ipt --noflush $ipt_arg
-	done
+
+        # Network Hooks for triggering supplementary actions on AppVM connect
+        if [ -d /rw/config/network-hooks.d ]; then
+            for hook in /rw/config/network-hooks.d/*
+            do
+                log debug "Executing network-hook $(basename "$hook")..."
+                do_without_error "${hook}" "${command}" "${vif}" "${addr}"
```
marmarek: Check if file is executable first.

network/vif-route-qubes
```diff
@@ -120,8 +130,7 @@ if [ "${ip}" ] ; then
 fi
```
marmarek: IMO the hook should be called even if IP isn't specified. In that case, the argument should be empty.

marmarek:
Also, please add info to https://www.qubes-os.org/doc/config-files/, including parameters, when hooks are called etc.

PR #161, opened 2019-06-09
thunar configuration

adrelanos:
Changes date style to iso format.
Disable network bookmark by default.
Disable thunar-volman by default.

### DIFF
--- a/misc/thunar.xml
+++ b/misc/thunar.xml
@@ -1,5 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
+<!-- Configuration for Thunar. -->
+<!-- Changes date style to iso format. -->
+<!-- Disable thumbnails. -->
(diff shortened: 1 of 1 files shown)


adrelanos:
Real solution would be https://github.com/QubesOS/qubes-issues/issues/1885 though which encompasses a lot more.

fepitre:
@adrelanos: yes maybe we could implement it directly? Initially, I put a really straightforward configuration but hardened one is much better.

adrelanos:
I am not working on this one anymore. Since https://github.com/QubesOS/qubes-issues/issues/1885 made progress, is the way to go, and this one would be conflicting with it, I am closing this one.

Feel free to come up with pull requests of your own, though. (Not an administrative statement. Just me stating I am not working on this specific solution anymore.)

PR #164, opened 2019-06-20
install jitterentropy-rngd by default

adrelanos:
Add `jitterentropy-rngd` as `Recommends:` to package `qubes-core-agent`.

https://github.com/QubesOS/qubes-issues/issues/4169

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -61,6 +61,7 @@ Recommends:
     gnome-terminal,
     gnome-themes-standard,
     haveged,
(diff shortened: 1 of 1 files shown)


marmarek:
Since we have meta-packages, I think a better place would be in https://github.com/QubesOS/qubes-meta-packages/

adrelanos:
Makes a lot sense.

https://github.com/QubesOS/qubes-meta-packages/pull/6

PR #166, opened 2019-07-11
Fix minimal

fepitre:


marmarek:
This will break jessie (debian 8) build, as `nftables` is not there.
In theory we still support jessie: https://www.qubes-os.org/doc/supported-versions/#templatevms
I'd be happy to drop support for oldoldstable debian version, but it needs to be done before breaking things...
In theory Debian jessie has security updates through Debian LTS [until end of June 2020](https://www.debian.org/releases/jessie/). But "regular" support already ended last year. @unman @andrewdavidwong do we have any policy written about it? I'd go with supporting only fully supported Debian versions, not LTS ones. Supporting 4+ years old Debian, which lack many modern packages (for example poor support for python3, many missing packages) is holding us back.

andrewdavidwong:
> This will break jessie (debian 8) build, as `nftables` is not there.
> In theory we still support jessie: https://www.qubes-os.org/doc/supported-versions/#templatevms
> I'd be happy to drop support for oldoldstable debian version, but it needs to be done before breaking things...
> In theory Debian jessie has security updates through Debian LTS [until end of June 2020](https://www.debian.org/releases/jessie/). But "regular" support already ended last year. @unman @andrewdavidwong do we have any policy written about it? I'd go with supporting only fully supported Debian versions, not LTS ones. Supporting 4+ years old Debian, which lack many modern packages (for example poor support for python3, many missing packages) is holding us back.

We discussed this approximately seven months ago:

https://groups.google.com/d/topic/qubes-devel/9_oMSncKj9g/discussion

But no, we do not have an official written policy about this.

unman:
> This will break jessie (debian 8) build, as nftables is not there.

It also seems like an unnecessary addition for stretch-minimal, no?

I favour supporting Debian oldstable through LTS, but not extending back to oldoldstable. (Which is what I said in the thread linked to.) If we need this stated as a policy I could add something to the Debian page, as well as clarifying [supported versions](https://www.qubes-os.org/doc/supported-versions/#templatevms)
For Ubuntu, we generally support only the last two LTS versions. (The most recent usually has to wait for Fedora debootstrap to catch up). This isnt stated anywhere.

marmarek:
@fepitre can you provide details what exactly breaks? `iptables` is already listed in `Depends:`, and AFAIK all relevant places here fallbacks to `iptables` if `nftables` is not there.

fepitre:
Ok. I will try to reproduce it again. It was breaking around port forwarding.

fepitre:
Finally, everything is working. I should have had network issues the first time. @marmarek: with respect to previous comments, I close the PR.

PR #167, opened 2019-07-14
use long options rather than short options

adrelanos:

This makes this easier to review, discuss, modify.


adrelanos:
Travis failed for unrelated reasons.

PR #168, opened 2019-07-14
Allow passwordless login for user "user" (when using 'sudo xl console').

adrelanos:

Running `sudo usermod --password '' user` in debian-10 TemplateVM allowed me to login as user "user" using `sudo xl console`. While untested, I am pretty sure this would work in the next template build too.

### DIFF
--- a/debian/qubes-core-agent.preinst
+++ b/debian/qubes-core-agent.preinst
@@ -51,6 +51,10 @@ if [ "$1" = "install" ] ; then
     }
     usermod -L -a --groups qubes user
 
+    ## Allow passwordless login for user "user" (when using 'sudo xl console').
+    ## https://github.com/QubesOS/qubes-issues/issues/1130
+    usermod --password '' user
+
     # --------------------------------------------------------------------------
     # Remove `mesg` from root/.profile?
     # --------------------------------------------------------------------------


debian/qubes-core-agent.preinst
```diff
@@ -51,6 +51,10 @@ if [ "$1" = "install" ] ; then
     }
     usermod -L -a --groups qubes user
 
+    ## Allow passwordless login for user "user" (when using 'sudo xl console').
+    ## https://github.com/QubesOS/qubes-issues/issues/1130
+    usermod --password '' user
```
marmarek: This is already done in upgrade case below. Maybe it should be unified, like that call below moved outside of `if`?

adrelanos: Yes, better. For simplicity and less bugs (at cost of some more needless output) I think it is always good to avoid using install and upgrade cases.

marmarek:
Hmm, when I think about it more, it may not be such a good idea to do it this way. It will enable passwordless login to `user` not only on `xl console`, but everywhere. User may decide to remove `qubes-core-agent-passwordless-root` package to have some isolation between accounts (for example sandboxing some services using dedicated users). In that case, having empty password for `user` would allow anyone to switch to `user` by calling `su - user`.
Note the problem apply much less for the `qubes-core-agent-passwordless-root` case, where user isolation is explicitly not relied upon.

Maybe better do it by configuring PAM to specifically allow passwordless user login on `hvc0` console?

marmarek:
Oh, I see you raised exactly the same concern in https://github.com/QubesOS/qubes-issues/issues/2695#issuecomment-518689427

adrelanos:
Answered here: https://github.com/QubesOS/qubes-issues/issues/2695#issuecomment-518733140

adrelanos:
> Ok, then...
> 
> Quote @marmarek [QubesOS/qubes-core-agent-linux#168 (comment)](https://github.com/QubesOS/qubes-core-agent-linux/pull/168#issuecomment-518694275)
> 
> > Maybe better do it by configuring PAM to specifically allow passwordless user login on hvc0 console?
> 
> Yes.

adrelanos:
This change is just doing what we are currently doing anyhow. But I can see how my approach can be confusing to break it more visibly before starting to fix it.

HW42:
> Maybe better do it by configuring PAM to specifically allow passwordless user login on hvc0 console?

Depending on how the exact result should look like, it might be easier to just run `agetty` with `--autologin user` (or `--autologin root`). (You probably also want to add `--login-pause`)

marmarek:
Autologin idea sounds much better and safer.

marmarek:

PR #169, opened 2019-07-14
Allow passwordless login for user "user" (when using 'sudo xl console').

adrelanos:
use long option names by default.

### DIFF
--- a/debian/qubes-core-agent-passwordless-root.preinst
+++ b/debian/qubes-core-agent-passwordless-root.preinst
@@ -34,7 +34,10 @@ set -e
 # the debian-policy package
 
 if [ "$1" = "install" ] ; then
-    usermod -p '' root
+    usermod --password '' root
+    ## Allow passwordless login for user "user" (when using 'sudo xl console').
+    ## https://github.com/QubesOS/qubes-issues/issues/1130
+    usermod --password '' user
 fi
 
 # dh_installdeb will replace this with shell code automatically


debian/qubes-core-agent-passwordless-root.preinst
```diff
@@ -34,7 +34,10 @@ set -e
 # the debian-policy package
 
 if [ "$1" = "install" ] ; then
-    usermod -p '' root
+    usermod --password '' root
+    ## Allow passwordless login for user "user" (when using 'sudo xl console').
+    ## https://github.com/QubesOS/qubes-issues/issues/1130
+    usermod --password '' user
```
marmarek: You take care about `user` in https://github.com/QubesOS/qubes-core-agent-linux/pull/168 already

adrelanos: qubes-core-agent-passwordless-root / https://github.com/QubesOS/qubes-core-agent-linux/pull/168 isn't installed everywhere. qubes-core-agent-passwordless-root will one day hopefully become an opt-in (or at least opt-out) package as per https://github.com/QubesOS/qubes-issues/issues/2695.

The end goal is to have a clean separation. qubes-core-agent should not not set up passwordless root (since hard to disable) but only qubes-core-agent-passwordless-root should set passwordless root up. To reach that goal, as preparation I wanted to use long parameters in preference of short parameters, comments (for better understanding) and consistency (using user and root everywhere) before actually fixing this.

marmarek: Yes, this is all fine. I pointed out that while you convert short options to long options, you also added setting empty **user** password, in qubes-core-agent-passwordless-**root** package. Setting user password belongs to `qubes-core-agent.postinst`

marmarek:

PR #172, opened 2019-07-20
dom0-updates: Quote arguments

m-v-b:
This pull request is intended to be merged together with https://github.com/QubesOS/qubes-core-admin-linux/pull/50

(This commit is applicable to the release4.0 branch as well.)


m-v-b:
I am in the process of fixing the `shellcheck` errors and will verify/test the resulting code.

m-v-b:
`shellcheck` errors should now be eliminated. (And they are apparently, according to the notification.)

I had to temporarily modify the script to force `yumdownloader` to be used (which also required removing the `--noplugins` option temporarily) so that all of the code paths could be verified. In any case, the behaviour of the script continues to be as expected.

PR #174, opened 2019-08-02
Allow creating TCP sockets between qubes

fepitre:


marmarek:
Some generic thoughts:
 - service name `qubes.Connect` IMO would make more sense
 - very much related to https://github.com/QubesOS/qubes-issues/issues/2148
 - UDP may be problematic, as it doesn't guarantee the same split into packets - do you have any example where UDP with this service would work?
 - command line tool `qvm-socket` is useful for ad-hoc connections, for persistent one systemd service would be more convenient (see https://github.com/QubesOS/qubes-core-agent-linux/blob/master/vm-systemd/qubes-updates-proxy-forwarder.socket and https://github.com/QubesOS/qubes-core-agent-linux/blob/master/vm-systemd/qubes-updates-proxy-forwarder%40.service which are in principle very similar thing)
 - it could be useful to specify target domain at qrexec policy level (and use `@default` target in `qrexec-client-vm` call), this way it will hide this info from sys-net (or wherever you're running it); you need to setup policy anyway, so it isn't more work

And few specific ones:
 - arguments: first protocol, then port
 - there are no multiple arguments: if you use `+` inside argument, it will be passed as a single argument `arg1+arg2`

fepitre:
> Some generic thoughts:
> 
>     * service name `qubes.Connect` IMO would make more sense
+1
>     * very much related to [QubesOS/qubes-issues#2148](https://github.com/QubesOS/qubes-issues/issues/2148)
I never saw this issue but yes indeed, it's very much related.
>     * UDP may be problematic, as it doesn't guarantee the same split into packets - do you have any example where UDP with this service would work?
Currently, I have no direct application in mind. I put UDP for no restricting things but maybe an application could be rsyslog.
>     * command line tool `qvm-socket` is useful for ad-hoc connections, for persistent one systemd service would be more convenient (see https://github.com/QubesOS/qubes-core-agent-linux/blob/master/vm-systemd/qubes-updates-proxy-forwarder.socket and https://github.com/QubesOS/qubes-core-agent-linux/blob/master/vm-systemd/qubes-updates-proxy-forwarder%40.service which are in principle very similar thing)
It makes sense.
> 
>     * it could be useful to specify target domain at qrexec policy level (and use `@default` target in `qrexec-client-vm` call), this way it will hide this info from sys-net (or wherever you're running it); you need to setup policy anyway, so it isn't more work
+1
> 
> And few specific ones:
> 
>     * arguments: first protocol, then port
> 
>     * there are no multiple arguments: if you use `+` inside argument, it will be passed as a single argument `arg1+arg2`
Ok thank you for the info.

marmarek:
I'm going to merge it now, but please also add documentation for this - preferably both man page (very simple one is fine), and somewhere in qubes-doc. Maybe [here](https://www.qubes-os.org/doc/firewall/#port-forwarding-to-a-qube-from-the-outside-world)?

PR #181, opened 2019-08-30
Disable caching in all apt operations

unman:
Closes QubesOS/qubes-issues#5266


unman:
> According to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=812111, it should be `Binary::apt::APT::Keep-Downloaded-Packages`. Does the current version work for you?

Sorry for delay in reply.

That setting is only to change the handling of `apt`, which has "Keep-Dpownloaded-Packages" OFF by default.
For other operations, like `apt-get` you need the setting I have here. (`apt-get` is best used in scripting and this is needed in our update templates operation)

unman:
@marmarek Can this be merged?

PR #183, opened 2019-09-07
Update python2 dependencies to python3 and clean deprecated requirements

fepitre:
QubesOS/qubes-issues#5289
QubesOS/qubes-issues#5297


marmarek:
Related: https://github.com/QubesOS/qubes-issues/issues/5290
Dropping setuptools package would make it worse. On Fedora it should be `python3-setuptools`, but not sure about centos.

fepitre:
Ok that's weird. Better keep it for CentOS and add it for Fedora

PR #187, opened 2019-10-06
qubes-download-dom0-updates: verify package signatures

illikainen:
This PR implements signature verification of downloaded RPMs in the UpdateVM.  The documentation on secure updates for dom0 mentions that the UpdateVM is considered [untrusted][1], and that dom0 is responsible for verifying package signatures.

However, I think it makes sense to designate a (semi-)trusted domU as the UpdateVM; especially with how easy it is to create new guests in Qubes.

By verifying package signatures on a (semi-)trusted domU, an attacker is put in a position of having to find an exploitable condition on **both** domU and dom0 in order to compromise the entire system through software updates.

Ideally, either domU or dom0 would use a different implementation for package verification (e.g. with a minimal RPM header parser and an alternative PGP implementation) to further mitigate the risk of exploitable bugs affecting both the guest and the host.  I'd be interested in implementing something along those lines if it's deemed desirable.

[1]: https://www.qubes-os.org/doc/dom0-secure-updates

### DIFF
--- a/misc/qubes-download-dom0-updates.sh
+++ b/misc/qubes-download-dom0-updates.sh
@@ -13,6 +13,8 @@ elif [ -f "$DOM0_UPDATES_DIR/etc/yum.conf" ]; then
 fi
 # DNF uses /etc/yum.repos.d, even when --installroot is specified
 OPTS="$OPTS --setopt=reposdir=$DOM0_UPDATES_DIR/etc/yum.repos.d"
+# DNF verifies signatures implicitly, but yumdownloader does not.
+SIGNATURE_REGEX=""
 PKGLIST=()
 YUM_ACTION=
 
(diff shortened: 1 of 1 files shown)


marmarek:
Thanks for the patch. This indeed makes very little sense using exactly the same implementation. Anyway, `dnf` already verify signatures and exit with non-zero code if it fails, even in downloadonly mode. So, there is no need for doing it again in this script.

illikainen:
I'd like to revisit this PR.

I checked `dnf` and it does indeed verify package signatures with `--downloadonly` (however I haven't been able to find any documentation that ensures it to be a feature to be relied upon in future releases of `dnf`).

I also checked `yumdownloader` and it does **not** verify package signatures; it'll happily download anything.  Hence my wanting to reopen discussion on this PR as `yumdownloader` is used on Debian-based UpdateVMs.

```sh
~$ cat yumdownloader/yum.conf
[main]
...

[slack]
name=slack
baseurl=https://packagecloud.io/slacktechnologies/slack/fedora/21/$basearch
gpgcheck=1
gpgkey = file:///etc/apt/trusted.gpg.d/qubes-archive-keyring.gpg

~$ yumdownloader --destdir=$(pwd)/yumdownloader/destdir --config=$(pwd)/yumdownloader/yum.conf slack
slack-4.0.2-0.1.fc21.x86_64.rpm                                                                                                                                   |  65 MB  00:01:34     

~$ echo $?
0
~$
```

All of RPM 4.13.x, GPG 2.1.x and libgcrypt 1.7.x seems EOL upstream, so it seems reasonable to assume that flaws affecting signature verification in dom0 might go unnoticed.  Especially since not all sec-related fixes are assigned CVEs.

marmarek:
Ok, for yumdownloader case indeed it makes sense.
Could you please update the patch to:
- use keys already present in rpmdb (see `rpmkeys --dbpath=... --checksig`)
- limit this to non-dnf case
- handle old `rpm --checksig` output format too (Debian stretch still has rpm 4.12, and this is where yumdownloader case would be used too)

illikainen:
b9a5fb7 handles both old and new versions of `rpm`.  It uses `dpkg` to compare `rpm` versions (which obv isn't ideal if there are plans for non-Debian-based UpdateVMs that uses `yumdownloader` in the future.)

Tested on Debian 9 and 10.

marmarek:
CentOS build failure unrelated.

PR #194, opened 2019-11-16
Raw USB devices may be used by the primary user

Osndok:
Some devices & programs (like Android's debug bridge `adb`) cannot presently be used in qubes without special knowledge and modification. This PR intends to make all USB devices usable by the system operator.

## Before: ##
```
crw-rw---- 1 root root 189, 4 Nov 16 09:55 /dev/bus/usb/001/005
```

## After: ##
```
crw-rw---- 1 root qubes 189, 4 Nov 16 09:55 /dev/bus/usb/001/005
```


marmarek:
Please see https://www.qubes-os.org/doc/code-signing/

PR #199, opened 2020-01-07
GetAppmenus: ensure right app directories

pwmarcz:
The script depends on XDG_DATA_DIRS environment variable
being set up correctly, which is not the case when it is
running under sudo. As a result, a post-install trigger
for apt could remove application entries from other sources
(Snap, Flatpak).

Fixes QubesOS/qubes-issues#5477.

### DIFF
--- a/qubes-rpc/qubes.GetAppmenus
+++ b/qubes-rpc/qubes.GetAppmenus
@@ -6,6 +6,15 @@
 #  - directories stored on /rw in case of "rw-only" persistence
 #  - nothing, otherwise
 
+# Reload scripts in /etc/profile.d/, in case they register additional
+# directories in XDG_DATA_DIRS and we forgot them
+# (e.g. because we are running under sudo).
+for i in /etc/profile.d/*.sh ; do
+    if [ -r "$i" ]; then
+        . "$i" >/dev/null
+    fi
+done
+
 if [ -z "$XDG_DATA_HOME" ]; then
     XDG_DATA_HOME="$HOME/.local/share"
 fi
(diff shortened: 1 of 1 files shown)


marmarek:
ShellCheck hates you (see failed travis job), please add a comment to ignore this warning (`shellcheck disable=SC1090`).

qubes-rpc/qubes.GetAppmenus
```diff
@@ -6,6 +6,16 @@
 #  - directories stored on /rw in case of "rw-only" persistence
 #  - nothing, otherwise
 
+# Reload scripts in /etc/profile.d/, in case they register additional
+# directories in XDG_DATA_DIRS and we forgot them
+# (e.g. because we are running under sudo).
+for i in /etc/profile.d/*.sh ; do
+    if [ -r "$i" ]; then
+	# shellcheck disable=SC1090
```
marmarek: Indentation is odd (3 instead of 4 spaces).

pwmarcz: Ouch, that's a tab, my editor misbehaved. I pushed an updated commit.

marmarek:
While at it, I think there is very similar issue with `~/.local/share/applications`. The script use `$HOME`, but if that's `/root`, then entries in `/home/user/.local/share/applications` will be lost. As you know from the other issue, hardcoding `/home/user` is a bad idea, but maybe we can use "default user" if the script is running as root?

pwmarcz:
How about this?

marmarek:
Lets see what shellcheck says, but should be fine.

PR #201, opened 2020-01-09
qubes-firewall: add anti-spoofing rules for connected machines

pwmarcz:
qubes-firewall will now blacklist IP addresses from all connected
machines on non-vif* interfaces. This prevents spoofing source or
target address on packets going over an upstream link, even if
a VM in question is powered off at the moment.

Depends on QubesOS/qubes-core-admin#303 which makes admin maintain
the list of IPs in qubesdb.

Fixes QubesOS/qubes-issues#5540.


pwmarcz:
Would that be `raw / prerouting` on input, and `raw / output` on output? I don't understand the output path very well.

marmarek:
Output I think needs to remain in mangle at lest for iptables (`OUTPUT` table is only about traffic generated locally, not forwarded from elsewhere).

pwmarcz:
I reproduced a test failure (`test_010_simple_proxyvm`) locally, and it looks like this fails for an empty list of connected IPs. Updated now.

PR #207, opened 2020-01-23
qubes-run-terminal: use gnome-terminal --wait, if supported

pwmarcz:
Fixes QubesOS/qubes-issues#4606.


pwmarcz:
Unfortunately this adds overhead of running `gnome-terminal` once to check capabilities.

I thought about doing it the other way: run `gnome-terminal --wait`, then if it fails, check if the failure was because it does not support `--wait` and if so, run it again. But it seems convoluted, and potentially will launch the process twice if something breaks.

PR #208, opened 2020-01-23
Add qubes.VMExec call, for running a single command

pwmarcz:
https://github.com/QubesOS/qubes-issues/issues/4850

### DIFF
--- a/qubesagent/test_vmexec.py
+++ b/qubesagent/test_vmexec.py
@@ -1,0 +1,52 @@
+from unittest import TestCase
+from unittest.mock import patch, call
+
+from qubesagent.vmexec import main, decode, DecodeError
+
+
+class TestVmExec(TestCase):
+    def test_00_decode_simple(self):
+        self.assertEqual(decode('echo+Hello'), [b'echo', b'Hello'])
+
+    def test_01_decode_empty(self):
+        self.assertEqual(decode('echo+'), [b'echo', b''])
+
+    def test_02_decode_escaping(self):
+        self.assertEqual(decode('echo+Hello-20world'),
+                         [b'echo', b'Hello world'])
+        self.assertEqual(decode('-0A-0D'),
+                         [b'\n\r'])
+        self.assertEqual(decode('-2Fbin-2Fls'),
+                         [b'/bin/ls'])
+        self.assertEqual(decode('ls+--la'),
+                         [b'ls', b'-la'])
+        self.assertEqual(decode('ls+---61'),
+                         [b'ls', b'-a'])
+        self.assertEqual(decode('ls+----help'),
+                         [b'ls', b'--help'])
+
+    def test_03_decode_errors(self):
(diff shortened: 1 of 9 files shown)


qubesagent/vmexec.py
```diff
@@ -0,0 +1,56 @@
...
+    pass
+
+
+def decode_part(part):
+    if not re.match(r'^[a-zA-Z0-9._-]*$', part):
+        raise DecodeError('illegal characters found')
+
+    part = part.encode('ascii')
+
+    part = part.replace(b'--', b'/')
```
marmarek: This could use a comment, especially why it's safe.

qubesagent/test_vmexec.py
```diff
@@ -0,0 +1,38 @@
...
+    def test_01_decode_empty(self):
+        self.assertEqual(decode('echo+'), [b'echo', b''])
+
+    def test_02_decode_escaping(self):
+        self.assertEqual(decode('echo+Hello-20world'),
+                         [b'echo', b'Hello world'])
+        self.assertEqual(decode('-0a-0D'),
+                         [b'\n\r'])
+        self.assertEqual(decode('ls+--la'),
+                         [b'ls', b'-la'])
```
marmarek: I'd add also a test like `ls+---61`

qubesagent/test_vmexec.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def test_03_decode_errors(self):
+        with self.assertRaises(DecodeError):
+            decode('illegal/slash')
+        with self.assertRaises(DecodeError):
+            decode('illegal-singledash')
```
marmarek: And here a test for incomplete encoded byte (at the end, and/or in the middle)

marmarek:
Also, besides commit messages, please also add issue reference(s) to the pull request description. Github the nicely list related PRs in the issue, including their state.

qubesagent/vmexec.py
pwmarcz: Right, it's actually not safe :( Things like `-2Fbin-2Fls` get converted to `/bin/ls` then replaced to `-bin-ls`. I found a better way.

pwmarcz:
Updated:
- changed the approach to less hacky, fixing the `/` case
- permitted only capital letters in hex
- stronger tests (slashes, three/four dashes, small letters, incomplete byte)

qubesagent/vmexec.py
```diff
@@ -0,0 +1,58 @@
...
+    pass
+
+
+def decode_part(part):
+    if not re.match(r'^[a-zA-Z0-9._-]*$', part):
+        raise DecodeError('illegal characters found')
+
+    part = part.encode('ascii')
+
+    escape_re = br'--|-([A-F0-9]{2})'
```
marmarek: re.compile?

pwmarcz: Added.

PR #209, opened 2020-01-24
firewall: drop INVALID state TCP packets

pwmarcz:
Packets detected as INVALID are ignored by NAT, so if they are not
dropped, packets with internal source IPs can leak to the outside
network.

See:

https://bugzilla.netfilter.org/show_bug.cgi?id=693
http://www.smythies.com/~doug/network/iptables_notes/

Fixes QubesOS/qubes-issues#5596.

### DIFF
--- a/network/ip6tables-enabled
+++ b/network/ip6tables-enabled
@@ -23,6 +23,7 @@ COMMIT
 :FORWARD DROP [0:0]
 :OUTPUT ACCEPT [0:0]
 :QBS-FORWARD - [0:0]
+-A INPUT -m state --state INVALID -j DROP
 -A INPUT -i lo -j ACCEPT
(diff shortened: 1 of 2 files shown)


pwmarcz:
Do you think we should do the same for `INPUT` chain? I understand this would block INVALID packets coming from the firewall machine itself, but these are not supposed to be NATed anyway.

marmarek:
> Do you think we should do the same for `INPUT` chain? I understand this would block INVALID packets coming from the firewall machine itself, but these are not supposed to be NATed anyway.

Yes, I think that's a good idea.

pwmarcz:
Okay, done. I verified that the rule is triggered for packets sent from other machine addressed to firewall machine - before, we would reject them with ICMP host prohibited, now we just drop them.

PR #210, opened 2020-01-27
Add override from gnome-terminal desktop shortcut

pwmarcz:
Fixes QubesOS/qubes-issues#2581.


pwmarcz:
Would also fix QubesOS/qubes-issues#3878, but gnome-terminal in Debian 9 is too old to support `--wait`.

marmarek:
I've added required travis setup to the master branch (pygtk is tricky to install there)

marmarek:
I wonder whether it's a good idea to backport this one to R4.0. On one hand, it fixes gnome terminal in DispVM. On the other hand, changes behavior `qvm-run --service vmname qubes.StartApp+org.gnome.Terminal` - and I know for sure some people's solutions depends on `qvm-run vmname gnome-terminal` returning immediately. Not sure about the variant with `qubes.StartApp`.

PR #213, opened 2020-02-05
fix https://github.com/QubesOS/qubes-issues/issues/5619

zaoqi:
python2 is being removed and python3-YYYY is being renamed to python-YYYY.


neowutran:
More modification are probably required.  `python2-nautilus` doesn't exist anymore, either this dependency is useless or should be replaced with its python3 version

marmarek:
FWIW all the code in master branch is already python3 compatible and should not require python2 at all. There is one exception about yum (which is python2-only), but that's only only for CentOS.

zaoqi:
updated

PR #219, opened 2020-02-09
Add README.md files describing qubes-core-agent-linux and its subprojects

aidecoe:


aidecoe:
@marmarek I will need your help with providing some of the READMEs, like the introduction for qubes-core-agent-linux in project root. Please also see TODO-s where I don't know how to explain something.

PR #220, opened 2020-02-14
init: skip journal creation in DispVMs

ambdroid:
This reduces writes and improves speed on HDDs. On my machine, this change shrinks the size of a freshly created DispVM (at time of "disp9999 has started" notification) from ~97MiB to ~16MiB.

Additionally, this fixes is_dispvm(), which was previously unused but would report that a DispVM is an AppVM.

### DIFF
--- a/init/setup-rwdev.sh
+++ b/init/setup-rwdev.sh
@@ -1,5 +1,9 @@
 #!/bin/sh
 
+# Source Qubes library.
+# shellcheck source=init/functions
+. /usr/lib/qubes/init/functions
+
 set -e
 
 dev=/dev/xvdb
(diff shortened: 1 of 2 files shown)


ambdroid:
Looks like the key is still propagating.

init/setup-rwdev.sh
```diff
@@ -16,7 +20,13 @@ if [ -e "$dev" ] ; then
...
         # the device is empty, create filesystem
         echo "Virgin boot of the VM: creating private.img filesystem on $dev" >&2
-        if ! content=$(mkfs.ext4 -m 0 -q "$dev" 2>&1) ; then
+        # journals are only useful on reboot, so don't write one in a DispVM
+        if is_dispvm ; then
+            journal="-O ^has_journal"
+        else
+            journal="-O has_journal"
+        fi
+        if ! content=$(mkfs.ext4 -m 0 -q "$journal" "$dev" 2>&1) ; then
```
marmarek: It is surprising that mkfs.ext4 accepts `-O ^has_journal` as a single argument too, but it does!

PR #221, opened 2020-02-16
Disable unnecessary services in Debian

unman:
Added services to preset file.
Closes QubesOS/qubes-issues#5206


unman:
@marmarek I suspect I need to check this some more, because it isnt clear to me what's starting the avahi service. The timers should be fine here.

PR #214, opened 2020-02-05
fix https://github.com/QubesOS/qubes-issues/issues/5619

zaoqi:
python2 is being removed and python3-YYYY is being renamed to python-YYYY.

PR #222, opened 2020-02-16
Archlinux: R4.0 python2->python3 and drop olivier key

neowutran:
Modification
===============
* Fix for https://github.com/QubesOS/qubes-issues/issues/5619, 4.0 only since a patch have already been applied to master (https://github.com/QubesOS/qubes-core-agent-linux/pull/213). It overlap with the fix suggested in https://github.com/QubesOS/qubes-core-agent-linux/pull/214 that haven't been merged (missing key) 
* Drop some now irrelevant constraint on packages version for archlinux
* Delete the GPG key of the old maintainer. A keyring package was being build with the GPG key of Olivier Medoc who was the maintainer of the Archlinux template. It is not the case anymore, so did modification in order to change the effect of keyring package. Now is doesn't add any new key, it just delete the Olivier Medoc key if present. If in the future a new maintainer want to provide packages:  add the public GPG key to PKGBUILD-keyring-keys  and set the fingerprint as trusted in PKGBUILD-keyring-trusted ("GPGFINDERPRINT:4:")

Quick checks that I did
===================
* Built the template successfully
* Installed the newly built template and verified that I can execute xterm correctly
* Updated the qubes packages in my old archlinux template and verified that I can execute xterm correctly

TODO
============
* I will do another pull request in master for the 2 points not related to issue 5619
* Maybe a good idea to add a section to the archlinux doc for more details about the keyring package


archlinux/PKGBUILD-qubes-repo-4.0.conf
```diff
@@ -1,2 +1,3 @@
 [qubes-r4.0]
-Server = http://olivier.medoc.free.fr/archlinux/current
+#Replace the server below with your own
+Server = https://neowutran.ovh/qubes/vm-archlinux/pkgs/
```
marmarek: Do you plan to maintain this repository? If so, please state it explicitly and provide a public key and also update info in template docs. Otherwise, at the very least comment out this line.

neowutran: Removed the references to my server

About https://github.com/QubesOS/qubes-core-agent-linux/pull/222#pullrequestreview-359468967 pretty much the first big project I try to contribute to, idk what are the good way to do a multi-branch pull request. But for this "master" pull request, I don't intend to do something more complex than cherry picking the commits in this one

PR #224, opened 2020-02-26
qubes.Restore: remove bashism found by new ShellCheck

rustybird:


fepitre:
See PR#223

rustybird:
Ah, right on. Force pushed, leaving only the ShellCheck fix.

PR #225, opened 2020-04-08
Hint about a possible update failure reason

strugee:
This feels ugly because it doesn't fix the underlying bug, but
it's incredibly hard to trace down the root cause in this
situation and it causes updates to fail which is pretty
critical. So this is probably better than nothing.

Ref QubesOS/qubes-issues#5052


marmarek:
IMO better than guessing, make the actual error easier to find. It should be written to `/var/lib/qubes/updates/errors` in dom0, so `qubes-dom0-update` script may check this file and show it to the user. If that still doesn't contain enough details, you can modify `qubes-receive-updates` script to save stderr of `qfile-dom0-unpacker` there too.

PR #232, opened 2020-05-28
Add permanent neighbor entries

DemiMarie:
This fixes IPv6 on OpenBSD HVMs, and probably fixes it on Arch HVMs (which run systemd-networkd).

### DIFF
--- a/network/setup-ip
+++ b/network/setup-ip
@@ -4,6 +4,14 @@
 # shellcheck disable=SC1091
 . /usr/lib/qubes/init/functions
 
+add_host_route () {
+    /sbin/ip -- route replace to unicast "$1" dev "$2" onlink scope host
+}
+
+add_default_route () {
+    /sbin/ip -- route replace to unicast default via "$1" dev "$2" onlink
+}
+
 configure_network() {
     local MAC="$1"
     local INTERFACE="$2"
@@ -17,20 +25,24 @@ configure_network() {
     local secondary_dns="${10}"
 
     /sbin/ifconfig "$INTERFACE" "$ip" netmask "$netmask"
+    /sbin/ip -- neighbour replace to "$ip" dev "$INTERFACE" \
+        lladdr "$MAC" nud permanent
     if [ -n "$ip6" ]; then
         /sbin/ifconfig "$INTERFACE" add "$ip6/$netmask6"
+        /sbin/ip -- neighbour replace to "$ip6" dev "$INTERFACE" \
+            lladdr "$MAC" nud permanent
     fi
     /sbin/ifconfig "$INTERFACE" up
 
     if [ -n "$gateway" ]; then
         /sbin/route add -host "$gateway" dev "$INTERFACE"
         if [ -n "$gateway6" ] && ! echo "$gateway6" | grep -q "^fe80:"; then
-            /sbin/route -6 add "$gateway6/$netmask6" dev "$INTERFACE"
+            add_route "$gateway6/$netmask6" "$INTERFACE"
         fi
         if ! qsvc disable-default-route ; then
-            /sbin/route add default gw "$gateway"
+            add_default_route "$gateway" "$INTERFACE"
             if [ -n "$gateway6" ]; then
-                /sbin/route -6 add default gw "$gateway6" dev "$INTERFACE"
+                add_default_route "$gateway6" "$INTERFACE"
             fi
         fi
     fi
--- a/network/setup-ip
+++ b/network/setup-ip
@@ -3,6 +3,7 @@
 # Source Qubes library.
 # shellcheck disable=SC1091
 . /usr/lib/qubes/init/functions
+set -uo pipefail
 
 add_host_route () {
     /sbin/ip -- route replace to unicast "$1" dev "$2" onlink scope host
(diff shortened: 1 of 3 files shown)


marmarek:
The change I think is ok, but please make this PR against master branch.

DemiMarie:
@marmarek please re-review

DemiMarie:
This is modified to include the antispoofing fixes, and to add the entries on eth0 as well.  At this point, it *should* be safe to disable ARP, but I would want to test more first.

network/setup-ip
```diff
@@ -17,20 +25,24 @@ configure_network() {
     local secondary_dns="${10}"
 
     /sbin/ifconfig "$INTERFACE" "$ip" netmask "$netmask"
+    /sbin/ip -- neighbour replace to "$ip" dev "$INTERFACE" \
+        lladdr "$MAC" nud permanent
```
marmarek, line 29: This adds VM's own IP+MAC to the neighbour table, I don't think you meant that. It should be rather gateway's IP+MAC. In theory there is an issue with that - you don't have gateway's MAC in this script. But in practice, _with Xen_, vif interface always has `fe:ff:ff:ff:ff:ff` MAC, at least on Linux (requires checking in MirageOS).

@hannesm can you confirm/deny if the above about MAC address on netback is true on MirageOS too?

network/vif-route-qubes
```diff
@@ -118,32 +118,51 @@ domid=${domid/.*/}
...
+        printf '%s\n' "*raw" \
+            "$iptables_cmd -i ${vif} ! -s ${addr} -j DROP" \
+            "$iptables_cmd ! -i vif+ -s ${addr} -j DROP" \
+            "COMMIT" | \
             ${cmdprefix} $ipt --noflush $ipt_arg
-
-        network_hooks "${command}" "${vif}" "${addr}"
+        if [[ "$command" = 'online' ]]; then
+            ip neighbour "${ipcmd}" to "${addr}" \
+                dev "${vif}" lladdr 00:16:3e:5e:6c:6c:00 nud permanent
```
marmarek, line 139: `6c` is duplicated. But also, this will break with custom MAC address (`mac` VM's property is writable). I do see `mac` property in xenstore for an interface. I'm not sure if it's already retrieved into some variable, but you can read it from xenstore (seems like vif-nat script does that, for example).

marmarek (changes requested):

network/vif-route-qubes
DemiMarie: Fixed

marmarek (approved):
This looks fine now, but I'll run it through openQA before merging (and first I need to sort out the current pile of failures there...)

network/setup-ip
hannesm: Thanks for the mention @marmarek. I think this is the case according to https://github.com/mirage/mirage-net-xen/pull/87 and https://github.com/mirage/qubes-mirage-firewall/pull/61 //cc @talex5.

network/vif-route-qubes
```diff
@@ -112,38 +112,58 @@ case "$command" in
...
-            log error "Cannot set IPv6 route to ${addr}, IPv6 disabled in the kernel"
-            continue
-        fi
-        ${cmdprefix} ip route "${ipcmd}" "${addr}" dev "${vif}" metric "$metric"
         if [[ "$addr" = *:* ]]; then
             ipt=ip6tables-restore
         else
             ipt=iptables-restore
         fi
-        echo -e "*raw\\n$iptables_cmd -i ${vif} ! -s ${addr} -j DROP\\nCOMMIT" | \
```
marmarek, line 137: Hmm, I wonder how it ever worked. This filters out link-local traffic, including neighbor discovery. Looking at my firewall, Indeed this rule dropped quite a lot of packets. Perhaps it worked only thanks to periodic "neighbor advertisement" sent?

DemiMarie: Perhaps.  That would explain why I kept having hangs with IPv6.  Adding the permanent neighbor entries should solve the problem, however.  A follow-up PR will drop all ARP and NDP traffic in both directions, since it is an attack vector and serves no purpose.

Ideally, Xen netfront and netback interfaces would be layer 3 point-to-point links, not layer 2 Ethernet links, but changing that will be a lot more work.

marmarek: Adding permanent neighbor is fine. But dropping ARP and NDP needs some more care - we do support running other systems, including without this package installed. This applies to Windows and other systems in HVM, but also to MirageOS or other unikernels (where you don't have stubdomain to handle ARP/NDP in). Anyway, that's discussion for the other PR.

network/vif-route-qubes
```diff
@@ -112,38 +112,58 @@ case "$command" in
...
+if [ "${ip}" ]; then
+    # If we've been given a list of IP addresses, then add routes from dom0 to
+    # the guest using those addresses.
+    for addr in ${ip};
+    do
+        if [[ "$addr" = *:* ]] && [[ "$ipv6_disabled" = 1 ]]; then
+            log error "Cannot set IPv6 route to ${addr}, IPv6 disabled in the kernel"
+            continue
+        fi
+        ${cmdprefix} ip route "${ipcmd}" "${addr}" dev "${vif}" metric "$metric" onlink
```
marmarek, line 163: With this PR applied, network doesn't work at all. I've got `Error: Invalid flags for nexthop - PERVASIVE and ONLINK can not be set.` error in `/var/log/xen/xen-hotplug.log` in sys-net and I think it comes from this line.

DemiMarie: Indeed it does.  Good catch!

marmarek:
>     * VmNetworking_fedora-32: [test_200_fake_ip_simple](https://openqa.qubes-os.org/tests/13864#step/VmNetworking_fedora-32/13)

I can't figure out what is wrong. 

Context for this test: it check the feature where the real VM's IP is hidden from it using NAT in a separate namespace (on the network-providing side). This way you can have multiple VMs that all thinks they have the same generic IP.

I launched it manually (the very same disk image) and pinged 192.168.123.45 from test-inst-vm1. Observations: the ping goes through test0 interface in test-inst-netvm1 (where 192.168.123.45 is set) response is generated but it never reaches test-inst-vm1.
Ping goes through:
1. test-inst-vm1: eth0
2. test-inst-netvm1: vif8.0 (in vif8.0-nat net namespace)
3. test-inst-netvm1: vif8.0-p (in vif8.0-nat net namespace)
4. test-inst-netvm1: vif8.0 (in main net namespace)
5. test-inst-netvm1: test0 (in main net namespace)

This one works.
Then pong goes back the same way in reversed order. I see it (with tcpdump) at step 3, but not 2.

I see no indication of any firewall rule blocking it (looking at counters), route entries also looks good in both namespaces.
Any idea?

DemiMarie:
The neighbor entry is in the initial network namespace, but it needs to be in `vif8.0-nat`.  Since `vif8.0-nat`’s veth endpoint has a MAC address other than `00:16:3e:5e:6c:00`, the frames that reach it are discarded.

marmarek:
Hmm, perhaps to simplify setup, the `vif8.0-nat`'s veth endpoint should have MAC forced to `00:16:3e:5e:6c:00`? It does mean duplicated MAC, but on different network interfaces so it shouldn't be a problem, at least in theory. After all, all `vif*` interfaces have the same MAC and it works.

marmarek:
(not really hardcoded `00:16:3e:5e:6c:00` but the one retrieved from xenstore)

marmarek:
I have an idea why some tests fail now. Without static neighbor entries, there was some ARP retry which gave more time for the network to start working. Now this delay is gone and if the network isn't up (on both ends) from the very beginning, the test fails.

DemiMarie:
That makes sense.  So the neighbor entries bring the network up faster, which exposes a race condition that was previously hidden.

PR #237, opened 2020-07-13
update-proxy-configs: handle Portage(Gentoo)

fepitre:


network/update-proxy-configs
```diff
@@ -144,3 +144,12 @@ fi
 if [ -e /etc/PackageKit/PackageKit.conf ]; then
     update_conf /etc/PackageKit/PackageKit.conf "ProxyHTTP=$PROXY_ADDR"
 fi
+
+# Portage (Gentoo)
+if [ -e /etc/portage/make.conf ]; then
+    if [ -n "$PROXY_ADDR" ]; then
+        update_conf /etc/portage/make.conf "all_proxy=$PROXY_ADDR"
+    else
+        sed -i '/all_proxy=/d' /etc/portage/make.conf
```
marmarek, line 153: You should still use `update_conf` for this (with empty argument, or a comment), to avoid messing with entries outside of `QUBES BEGIN/END` section.

fepitre: I inserted a comment. Empty value is not working.

PR #239, opened 2020-07-24
Drop legacy xen entry in fstab

fepitre:


vm-systemd/75-qubes-vm.preset
```diff
@@ -100,11 +100,12 @@ enable qubes-iptables.service
 enable qubes-updates-proxy-forwarder.socket
 enable haveged.service
 enable chronyd.service
-enable xendriverdomain.service
```
marmarek, line 103: This is still necessary.

vm-systemd/xendriverdomain.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Xen driver domain device daemon
+ConditionVirtualization=xen
+
+[Service]
+Type=forking
+ExecStart=@sbindir@/xl devd
```
marmarek, line 7: This should be `/usr/sbin/xl devd`...

fepitre: Yes

PR #243, opened 2020-08-22
Workaround for gpg not resolving key servers used behing proxy

fepitre:
See QubesOS/qubes-issues#6013


fepitre:
I'm closing it currently as there work in progress for using alternative to `gpg` into the refresh of keys.

PR #245, opened 2020-08-25
Skip IGD when unbinding device drivers on suspend

cytadela8:
Needed and potentially sufficient for sys-gui-gpu suspend.

Part of: QubesOS/qubes-issues#2618

### DIFF
--- a/qubes-rpc/prepare-suspend
+++ b/qubes-rpc/prepare-suspend
@@ -38,9 +38,15 @@ if [ x"$action" = x"suspend" ]; then
     # detach all drivers from PCI devices (the real ones, not emulated by qemu)
     echo -n > /var/run/qubes-suspend-pci-devs-detached
     for dev_path in /sys/bus/pci/devices/*; do
-        # skip qemu emulated devs
         subsystem_vendor=$(cat "$dev_path/subsystem_vendor")
+        vendor=$(cat "$dev_path/vendor")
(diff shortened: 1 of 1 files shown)


qubes-rpc/prepare-suspend
```diff
@@ -38,11 +38,16 @@ if [ x"$action" = x"suspend" ]; then
...
     for dev_path in /sys/bus/pci/devices/*; do
-        # skip qemu emulated devs
         subsystem_vendor=$(cat "$dev_path/subsystem_vendor")
+        class=$(cat "$dev_path/class")
+        # skip qemu emulated devs
         if [ "$subsystem_vendor" = "0x1af4" ] || [ "$subsystem_vendor" = "0x5853" ]; then
             continue
         fi
+        # skip Intel Graphics Device
+        if [ "$subsystem_vendor" = "0x1028" ] && [ "$class" = "0x030000" ]; then
```
marmarek, line 48: This looks suspicious - vendor 0x1028 is Dell - looks like the laptop manufacturer, not graphics card itself. This should check "vendor", not "subsystem_vendor".

marmarek (changes requested):

qubes-rpc/prepare-suspend
cytadela8: Fixed

marmarek (approved):

PR #253, opened 2020-10-18
Fix archlinux packaging

fepitre:
https://github.com/QubesOS/qubes-issues/issues/6028

### DIFF
--- a/archlinux/PKGBUILD
+++ b/archlinux/PKGBUILD
@@ -1,6 +1,6 @@
 #!/bin/bash
 # shellcheck disable=SC2034
-pkgname=(qubes-vm-core qubes-vm-networking qubes-vm-keyring)
+pkgname=(qubes-vm-core qubes-vm-networking qubes-vm-keyring qubes-vm-passwordless-root)
 pkgver=$(cat version)
 pkgrel=15
 epoch=
@@ -31,7 +31,7 @@ noextract=()
 md5sums=(SKIP)
 
 build() {
-    for source in autostart-dropins applications-dropins app-menu qubes-rpc misc Makefile vm-init.d vm-systemd network init version doc setup.py qubesagent post-install.d; do
+    for source in autostart-dropins applications-dropins app-menu qubes-rpc misc passwordless-root Makefile vm-init.d vm-systemd network init version doc setup.py qubesagent post-install.d; do
         # shellcheck disable=SC2154
         (ln -s "$srcdir/../$source" "$srcdir/$source")
     done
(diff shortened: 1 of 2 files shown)


archlinux/PKGBUILD
```diff
@@ -67,22 +66,20 @@ package_qubes-vm-core() {
...
     make -C qubes-rpc/nautilus DESTDIR="$pkgdir" install
     make -C qubes-rpc/thunar DESTDIR="$pkgdir" install
 
     # Remove things non wanted in archlinux
-    rm -r "$pkgdir/etc/yum"*
-    rm -r "$pkgdir/etc/dnf"*
-    rm -r "$pkgdir/etc/init.d"
+    rm -rf "$pkgdir/etc/yum"*
+    rm -rf "$pkgdir/etc/dnf"*
+    rm -rf "$pkgdir/etc/init.d"
```
marmarek, line 80: Are those installed at all? Since Makefile rework I think they aren't and those rm can be dropped.

archlinux/PKGBUILD
```diff
@@ -67,22 +66,20 @@ package_qubes-vm-core() {
...
     # Remove fedora specific scripts
-    rm "$pkgdir/etc/fstab"
+    rm -rf "$pkgdir/etc/fstab"
```
marmarek, line 82: Same here. But also -r is not needed for a file.
BTW, isn't lack of the fstab here the reason for / not remounted rw? Or is fstab created somewhere else?

marmarek (changes requested):
shellckeck complains about $pkgdir. While it looks like a false positive, please mute this warning somehow. Preferably in a way to still detect other unset variables. Maybe add something like:
`[ -n "${pkgdir:-}" ] || exit 1` at the beginning of relevant function(s)?

archlinux/PKGBUILD
fepitre: The archlinux fstab is setup here: https://github.com/QubesOS/qubes-builder-archlinux/blob/master/scripts/04_install_qubes.sh#L59 so unless my manual symlink it's broken.

archlinux/PKGBUILD
fepitre: Yes I've dropped it. I need to remove that too in gentoo ebuild.

fepitre:
> shellckeck complains about $pkgdir. While it looks like a false positive, please mute this warning somehow. Preferably in a way to still detect other unset variables. Maybe add something like:
> `[ -n "${pkgdir:-}" ] || exit 1` at the beginning of relevant function(s)?

Yes looking at some PKGBUILD example it seems to simply disable the check. Normally, only pkgdir variable warning is disabled. I've tried to add another one variable implicitly and the warning appears.

marmarek (approved):

PR #254, opened 2020-10-23
archlinux: improve pacman proxy implementation

ejose19:


ejose19:
@fepitre Let me know what you think

fepitre:
@ejose19 It looks good to me. I would need to test it too.

network/update-proxy-configs
```diff
@@ -119,21 +119,26 @@ fi
 # Pacman (archlinux) also
 if [ -d /etc/pacman.d ]; then
     if [ -n "$PROXY_ADDR" ]; then
-        cat > /etc/pacman.d/01-qubes-proxy.conf <<EOF
+        mkdir -p /usr/lib/qubes/bin
+        cat > /usr/lib/qubes/bin/pacman <<EOF
```
marmarek, line 123: If going with a non-standard path anyway, can we please not dynamically modify content of `/usr`? Better put it in `/var` or even `/run`

network/update-proxy-configs
```diff
@@ -119,21 +119,26 @@ fi
...
+        chmod +x /usr/lib/qubes/bin/pacman
+        cat > /etc/profile.d/qubes-proxy.sh << EOF
+### This file is automatically generated by Qubes ($0 script).
+### All modifications here will be lost.
+export PATH=/usr/lib/qubes/bin:\$PATH
 EOF
     else
-        rm -r /etc/pacman.d/01-qubes-proxy.conf
+        rm -r /usr/lib/qubes/bin/pacman
+        rm -r /etc/profile.d/qubes-proxy.sh
```
marmarek, line 137: Both wants a `-f` instead of `-r`.

network/update-proxy-configs
```diff
@@ -119,21 +119,26 @@ fi
...
+#!/bin/bash
 ### This file is automatically generated by Qubes ($0 script).
 ### All modifications here will be lost.
-### If you want to override some of this settings, create another file under 
-### /etc/pacman.d
-
-[options]
-# Use Qubes Update Proxy
-XferCommand = /usr/bin/env ALL_PROXY=$PROXY_ADDR /usr/bin/curl -C - -f -o %o %u
+ALL_PROXY=$PROXY_ADDR /usr/bin/pacman \$@
```
marmarek, line 127: Add `exec` there, to avoid extra process.

marmarek (changes requested):

ejose19:
@marmarek Should be good now. 

P.S. No idea why the code-signing check still fails, I can run ```gpg --keyserver pool.sks-keyservers.net --recv-key 09BEDF7CC137762C``` on a dispvm and it returns the key.

network/update-proxy-configs
```diff
@@ -119,21 +119,26 @@ fi
...
+#!/bin/bash
 ### This file is automatically generated by Qubes ($0 script).
 ### All modifications here will be lost.
-### If you want to override some of this settings, create another file under 
-### /etc/pacman.d
-
-[options]
-# Use Qubes Update Proxy
-XferCommand = /usr/bin/env ALL_PROXY=$PROXY_ADDR /usr/bin/curl -C - -f -o %o %u
+exec env ALL_PROXY=$PROXY_ADDR /usr/bin/pacman \$@
```
marmarek, line 127: ah, one more thing: it should be `"$@"`, otherwise arguments with spaces won't be handled correctly.

ejose19: Can you give me an example of such command so next time I can do a better testing? I tried with ```pacman -S -y -u --color always``` and it worked without issues.

marmarek: It's about arguments with spaces inside, for example:
`pacman --print --print-format "package %n version %v" -U`
(not sure if I got the syntax right, written based on pacman man page)

marmarek (approved):

network/update-proxy-configs
ejose19: I tried your command and it failed without the double quotes, thanks.

ejose19:
https://github.com/QubesOS/qubes-issues/issues/5585 Can be closed now as this is merged

PR #255, opened 2020-10-28
Overwrite .rpmdb for debian updatevm

icequbes1:
Resolves issue where the dom0 rpm database does not get used on
successive calls to qubes-dom0-update for debian updatevms.

Also resolves "cannot remove .rpmdbold.####" occurrences.

qubesos/qubes-issues#6124


marmarek:
Is there any place I can get your gpg pubkey? keyservers are not reliable anymore :/

marmarek (approved):

PR #268, opened 2020-12-07
Various network fixes

DemiMarie:
See individual commit messages for details.


network/vif-route-qubes
```diff
@@ -47,6 +47,14 @@ network_hooks() {
     fi
 }
 
+conntrack_purge () {
+    local n output deleted msg
+    n='(0|[1-9][0-9]*)' msg='flow entries have been deleted\.$'
+    deleted="^conntrack v$n\\.$n\\.$n \\(conntrack-tools\\): $n $msg"
+    output=$(LC_ALL=C exec conntrack -D "$@" 2>&1 >/dev/null) || :
```
marmarek, line 54: This requires adding a dependency on `conntrack-tools`, to all present packaging. Otherwise the command is not available, which prevents any `vif*` interface from coming up...

marmarek (changes requested):

network/vif-route-qubes
DemiMarie: Oops!  Fixed, at least for Arch, Debian, and Fedora.

marmarek:
> * system_tests_network
>   
>   * VmNetworking_fedora-32: [test_020_simple_proxyvm_nm](https://openqa.qubes-os.org/tests/14485#step/VmNetworking_fedora-32/3)
> 
> * system_tests_network_ipv6
>   
>   * VmIPv6Networking_fedora-32: [test_020_simple_proxyvm_nm](https://openqa.qubes-os.org/tests/14496#step/VmIPv6Networking_fedora-32/3)
>   * VmIPv6Networking_fedora-32: [test_520_ipv6_simple_proxyvm_nm](https://openqa.qubes-os.org/tests/14496#step/VmIPv6Networking_fedora-32/23)

Curiously, `simple_proxyvm_nm` test fails all three times it is called on fedora-32 based VM, but it passes on debian-10 and *fedora-32-xfce*! Investigating...
(because of the other issue in vif-route-qubes, openQA failed to upload logs...)

marmarek (approved):

PR #271, opened 2020-12-14
Install symlinks for keys in /etc

DemiMarie:
This makes `qubes-dom0-update` work with metadata verification.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,7 +160,7 @@ install-netvm:
 	install -D network/qubes-setup-dnat-to-ns $(DESTDIR)$(LIBDIR)/qubes/qubes-setup-dnat-to-ns
 
 	install -d $(DESTDIR)/etc/dhclient.d
-	ln -s /usr/lib/qubes/qubes-setup-dnat-to-ns $(DESTDIR)/etc/dhclient.d/qubes-setup-dnat-to-ns.sh
+	ln -s ../../usr/lib/qubes/qubes-setup-dnat-to-ns $(DESTDIR)/etc/dhclient.d/qubes-setup-dnat-to-ns.sh
 
 	install -D network/vif-route-qubes $(DESTDIR)/etc/xen/scripts/vif-route-qubes
 	install -D network/vif-qubes-nat.sh $(DESTDIR)/etc/xen/scripts/vif-qubes-nat.sh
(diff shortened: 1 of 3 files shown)


marmarek (changes requested):
I think if we want Debian-based UpdateVM to verify the metadata (which will be possible with dnf there), we need to install the keys in that package too (currently they are not included under `/etc/pki/rpm-gpg`).

An alternative approach could be make `qubes-dom0-update` send the keys from dom0 (under expected paths) - which could be cleaner solution, since the keys and their paths are dictated by the repository config - which we do send from dom0.

DemiMarie:
Pipeline retry

DemiMarie:
PipelineRetry

DemiMarie:
> An alternative approach could be make `qubes-dom0-update` send the keys from dom0 (under expected paths) - which could be cleaner solution, since the keys and their paths are dictated by the repository config - which we do send from dom0.

Would that be under `/etc/pki/rpm-gpg`?  That seems to have a high risk of breakage.

marmarek:
We copy rest of the dnf configuration (including yum.repos.d) into `/var/lib/qubes/dom0-updates`, so I'd assume it should go into `/var/lib/qubes/dom0-updates/etc/pki/rpm-gpg`, but I'm not sure whether will it work (specifically, if key paths are affected by the `--installroot` option or not).

DemiMarie:
> We copy rest of the dnf configuration (including yum.repos.d) into `/var/lib/qubes/dom0-updates`, so I'd assume it should go into `/var/lib/qubes/dom0-updates/etc/pki/rpm-gpg`, but I'm not sure whether will it work (specifically, if key paths are affected by the `--installroot` option or not).

I highly doubt it.  In particular, there would otherwise be no way to install into an empty chroot.

marmarek:
Indeed, you are right (just tested it). But we could patch the key path(s) using `qubes-download-dom0-updates.sh` script, though. Depending on a specific VM configuration files for downloading dom0 updates is a layering violation. This is also why we use dnf.conf sent by dom0 and repositories config from dom0 (even though dom0 doesn't really make use for real repository configs).

DemiMarie:
https://github.com/QubesOS/qubes-core-admin-linux/pull/68/commits/79fa7e43b239769d7039eda38d8bf6176a1b1f66 modifies `qubes-dom0-update` to upload the keys and patch the repository files.

DemiMarie:
Closing in favor of https://github.com/QubesOS/qubes-core-admin-linux/pull/68, which handles this more generically.

PR #260, opened 2020-11-12
Metadata is now signed

DemiMarie:
Fixes https://github.com/QubesOS/qubes-issues/issues/6177

PR #276, opened 2020-12-22
Always pass ‘-y’ to dnf

DemiMarie:
DNF should never be used unattended without ‘-y’.


DemiMarie:
Closing in favor of #260 which now includes this.

PR #277, opened 2020-12-24
Turn off middle mouse paste in gnome-terminal

DemiMarie:
It is *very* easy to trigger this by accident, and this can easily
result in system compromise if one is not careful.


DemiMarie:
PipelineRetry

marmarek:
This is not a place for your personal preferred setting (see https://www.qubes-os.org/faq/#could-you-please-make-my-preference-the-default). I recognize _your_ workflow may be greatly improved/hardened with such setting, and you are free to configure it in your template(s). It is also fine to document a set of helpful settings that some may find useful (the place for that is at https://github.com/Qubes-Community/Contents/). But definitely not put into a mandatory package.

PR #280, opened 2020-12-29
Ignore more options of qubes-dom0-update

DemiMarie:


marmarek (approved):

package-managers/qubes-download-dom0-updates.sh
```diff
@@ -22,7 +22,7 @@ export LC_ALL=C
 
 while [ -n "$1" ]; do
     case "$1" in
-        --doit)
+        --doit|--force-xen-upgrade|--console|--show-output)
```
marmarek, line 25: @DemiMarie  When now testing related code, it seems better to simply don't pass those options (affecting only dom0 side behavior) to the VM at all. This will avoid the need for updated core-agent-linux package in a VM to use updated qubes-dom0-update tool.

DemiMarie: Good point.  That will require some refactoring on the dom0 side.

PR #281, opened 2020-12-29
Avoid deprecated /var/run directory

DemiMarie:
It causes systemd to emit warnings.


marmarek:
We have this issue in more than one repo, can you create a tracking issue and link this PR to it?

DemiMarie:
Done: https://github.com/QubesOS/qubes-issues/issues/6315.  It’s `P: minor`, but as it is so trivial it would be a shame not to fix it.

marmarek (approved):

PR #286, opened 2021-01-05
archlinux: pin PKGBUILD to python3.X major version as new python vers…

ptitdoc:
…ion will break the API.

Should also be applied to branch 4.1 if possible to cherry-pick.

### DIFF
--- a/archlinux/PKGBUILD
+++ b/archlinux/PKGBUILD
@@ -1,5 +1,3 @@
-#!/bin/bash
-# shellcheck disable=SC2034
 pkgname=(qubes-vm-core qubes-vm-networking qubes-vm-keyring)
 pkgver=$(cat version)
 pkgrel=15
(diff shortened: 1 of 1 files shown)


marmarek:
Is it the right solution for this problem? This feels like something that should have a generic solution, not a manual version limitation. How other packages are doing this?
Solutions I've seen in other distributions:
 - use versioned shebang in python scripts (`#!/usr/bin/python3.8`) - CentOS has build scripts to do that automatically
 - a tools for automatic rebuilds of all python packages on major python update (Gentoo)

marmarek:
Related thread: https://bbs.archlinux.org/viewtopic.php?id=261341 it seems more people have this issue...

ptitdoc:
It is more the gentoo way. They rebuild and repackage everything upon new python version and /usr/lib/python3.7 is now empty except for qubes stuff that has not been rebuilt. Similar than what they do for glibc updates. Basically they don't care because users should upgrade all the time.This is however problematic for third party packages / repository . Here the goal is more to ensure pacman avoid updating to an unusuable qube template (this breaks gui agent). But the main problem is just ensuring that we repackage. Maybe I can try adjust to force to the existing python major version instead of having a static value. And there is anyway a kind of chicken and egg problem. Because when there is a python update I cannot update it to rebuild the qubesos package

marmarek:
I see... ok, I'm going to merge this for now, later we can improve (to have a dynamic, build-time value picked). Similar problem applies to other our packages too (for example core-qubesdb), but I guess just this one dependency will be enough to prevent breaking upgrade.

PR #287, opened 2021-01-08
qubes-early-vm-config.service: Wants=network-pre.target

rustybird:
The unit on the Before= side of network-pre.target also has to pull it
in as a dependency:

https://www.freedesktop.org/software/systemd/man/systemd.special.html#network-pre.target

Fixes QubesOS/qubes-issues#5570


marmarek:
> ##
> 
> No failures!

This doesn't happen. Actually, all the tests failed, because test runner run out of disk space ...

PR #289, opened 2021-01-11
sudo isn’t always built with SELinux support

DemiMarie:
Red Hat- and Debian- derived distributions support SELinux, and so their
sudo packages are built with SELinux support.  However, other
distributions (notably Arch) build sudo without SELinux.  Such sudo
builds will fail to parse the `ROLE=unconfined_r TYPE=unconfined_t`
string added in 0fac1aa45c677d82b413d429dae37f51ad1d67f4.  They *can*
parse `role=unconfined_r, type=unconfined_t` in `Defaults`, but that
causes problems on some Fedora 33 systems if SELinux is turned off and
the root account is locked.

To solve both of these problems at once, we install a different
`/etc/sudoers.d/qubes` file depending on the distribution.  As a
heuristic, we use the presents of `/etc/redhat-release` or
`/etc/debian_version`.  If either is present, sudo probably supports
SELinux, and we should include the corresponding entries.  If both are
missing, then we shouldn’t risk it.  The `qubes.sudoers` file in the git
repository includes the full file (with SELinux); we use `sed` to strip
the SELinux portion when needed.


marmarek (approved):

ejose19:
A release of ```qubes-core-agent-linux``` for v4.0 should be made soon, or else many people will start posting that their sudo access broke (I just updated and noticed the error, removing selinux related options from ```/etc/sudoers.d/qubes``` (as this PR) solved the issue.

PR #290, opened 2021-01-12
Open certain applications in disposable VMs

DemiMarie:
See individual commit messages for details.
Fixes  QubesOS/qubes-issues#6877


qubes-rpc/qubes.StartApp
```diff
@@ -1,25 +1,50 @@
...
+    paths_to_try = [xdg_data_home] + xdg_data_dirs
+    qubes_db = qubesdb.QubesDB()
+    if qubes_db.read('/qubes-service/app-dispvm.' + arg) is not None:
+        try:
+            del os.environ['XDG_DATA_HOME']
+        except KeyError:
+            pass
+        os.environ['XDG_DATA_DIRS'] = '/usr/share/qubes/mime'
+    for path in paths_to_try:
+        desktop_file = os.path.join(path, app_name)
```
marmarek, line 38: This alone won't be enough, you need to check also subdirectories - see `find ... -type d` call in the original code. This applies to some KDE applications.
On a second thought, we might have encoded application ID wrong - see https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#desktop-file-id, specifically `turn '/' into '-'` part. Currently we take only the last path part. 
And also, the specification is explicitly ambiguous: `If both foo-bar.desktop and foo/bar.desktop exist, it is undefined which is selected`.

qubes-rpc/qubes.StartApp
```diff
@@ -1,25 +1,50 @@
...
+                xdg_data_dirs.startswith(':') or \
+                xdg_data_dirs.endswith(':'):
+            print('Empty entries in XDG_DATA_DIRS not allowed')
+            sys.exit(1)
+        xdg_data_dirs = xdg_data_dirs.split(':')
+    paths_to_try = [xdg_data_home] + xdg_data_dirs
+    qubes_db = qubesdb.QubesDB()
+    if qubes_db.read('/qubes-service/app-dispvm.' + arg) is not None:
+        try:
+            del os.environ['XDG_DATA_HOME']
```
marmarek, line 33: As discussed elsewhere, according to the spec, this will make an application use the default value. But there is also another concern - applications using `$XDG_DATA_HOME` for other things than just mime database. I don't see any obvious directories in _my_ `~/.local/share` that would be affected (be used by an application I'd like to use this feature on), but it is still a risk, and a reason to keep this (for now) as per-application feature.
In fact, similar risk applies to `$XDG_DATA_DIRS`, for example `/usr/share/icons`, or `/usr/share/fonts` (not sure if the latter is sensitive to `$XDG_DATA_DIRS`). For this, probably some symlinks will be enough.

qubes-rpc/qubes.StartApp
```diff
@@ -1,25 +1,50 @@
...
+    try:
+        (arg0, arg,) = argv
+    except ValueError:
+        print('This service requires an argument', file=sys.stderr)
+        sys.exit(1)
+    app_name = os.path.join('applications', arg + '.desktop')
+    try:
+        xdg_data_home = os.environ['XDG_DATA_HOME']
+    except KeyError:
+        xdg_data_home = os.path.join(pwd.getpwuid(os.getuid()).pw_dir, '.local/share')
```
marmarek, line 17: I believe canonical way for doing this is `os.path.expanduser('~/.local/share')`, possibly with a check if `$HOME` is set.

marmarek (changes requested):

marmarek:
> This
> ensures that a desktop file that exists but cannot be accessed is not
> silently skipped.

I don't think that was an issue in the old code - see `-f` check in the old code - if that test passed, it ended up with `exec qubes-desktop-run`, so it never get back to the loop (and if file was unreadable, it doesn't undo `exec`).

marmarek:
See also CI - you miss `shared-mime-info` (providing `update-mime-database`) build dependency.

marmarek (changes requested):
It isn't really a change request in this PR, but I'm marking it as such, to note something is required before merging it.

qubes-rpc/qubes.StartApp
```diff
@@ -1,25 +1,55 @@
...
+            del os.environ['XDG_DATA_HOME']
+        except KeyError:
+            pass
+        os.environ['XDG_DATA_DIRS'] = '/usr/share/qubes/mime'
+    for path in paths_to_try:
+        subpath = os.path.join(path, 'applications')
+        for (raw_volume, directories, files) in os.walk(subpath):
+            volume = raw_volume[len(subpath):]
+            if volume:
+                volume = volume[1:].replace('/', '-') + '-'
```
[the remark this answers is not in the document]
DemiMarie: Fixed by 4d8fe0a9322e92839cbdf0ab86972b3f5fa5a51b.

marmarek:
Am I missing something or this PR lacks the key part: actually calling qvm-open-in-dvm from somewhere?

DemiMarie:
> Am I missing something or this PR lacks the key part: actually calling qvm-open-in-dvm from somewhere?

Indeed it does, oops!

qubes-rpc/qvm-open-in-dvm
```diff
@@ -20,10 +20,19 @@
...
+if ! [ $# = 1 ]; then
 	echo "Usage: $0 [--view-only] filename"
 	exit 1
 fi
-
-# shellcheck disable=SC2016
-exec qvm-open-in-vm '$dispvm' "$@"
+case $1 in
+/*) exec qvm-open-in-vm '@dispvm' $view_only "$1";;
+*) exec qvm-open-in-vm '@dispvm' $view_only "$1";;
```
marmarek, line 37: Why this? Both cases do exactly the same thing.

DemiMarie: That was meant to be `"./$1"`, will fix.

qubes-rpc/qubes.StartApp
```diff
@@ -1,25 +1,55 @@
...
+    except KeyError:
+        xdg_data_dirs = ['/usr/local/share','/usr/share']
+    else:
+        if '::' in xdg_data_dirs or \
+                xdg_data_dirs.startswith(':') or \
+                xdg_data_dirs.endswith(':'):
+            print('Empty entries in XDG_DATA_DIRS not allowed')
+            sys.exit(1)
+        xdg_data_dirs = xdg_data_dirs.split(':')
+    paths_to_try = [xdg_data_home] + xdg_data_dirs
```
marmarek, line 30: When having this in python, we can use already existing implementation: this whole thing can be replaced by:
```python
from xdg.BaseDirectory import xdg_data_dirs
```
(it includes xdg_data_home at the beginning already)

misc/applications/dispvm.desktop
```diff
@@ -0,0 +1,7 @@
+[Desktop Entry]
+Version=1.0
+Name=QubesOS Open In DisposableVM
```
marmarek, line 3: Maybe call it "View in DisposableVM"? To be distinct from https://github.com/QubesOS/qubes-core-agent-linux/pull/310
And also please name the file more descriptively, maybe `qvm-view-in-dvm`?
But if `--view-only` is not necessary (IMO is), then that desktop file could be re-used.

misc/mime/packages/qubes.org.xml
```diff
@@ -0,0 +1,380 @@
+<?xml version="1.0" encoding="utf-8"?>
```
marmarek, line 1: Shouldn't the filename be `qubes-os.org.xml`? Or dash is not allowed?

marmarek (changes requested):

misc/applications/dispvm.desktop
marmarek: > But if `--view-only` is not necessary (IMO is), then that desktop file could be re-used.

This is actually not that obvious. I was thinking about the mail attachment case only, but with this being a generic feature, it does makes sense to preserve modifications made in the DispVM. And as for the attachment case, I believe Thunderbird will remove the temporarily saved file anyway (needs checking), so allowing modification shouldn't really matter.

marmarek:
Have you tested it with Thunderbird? Or maybe some other application?

Here are my results, done on Fedora 33 VM:
 - with Nautilus, it doesn't work at all (the file is opened locally) - likely because Nautilus opens with dbus activation, so the environment isn't passed to it this way. This is the whole class of applications, I propose to focus on this later.
 - with Thunderbird, it also doesn't work, but in a different way:
   - text/plain attachment is detected as "Text File" (contrary to "plain text document" with normal mime database), and the default action is save instead of open.
   - There is no "View in DisposableVM" application to choose (there is just "Browse" button to choose an app path).
   - Furthermore, when accepting "Save File" action, Thunderbird crashes with assertion failure, apparently because PNG files are not recognized. This is awkward...

Actual abort message:
```
Gtk:ERROR:gtkiconhelper.c:494:ensure_surface_for_gicon: assertion failed (error == NULL): Failed to load /org/gtk/libgtk/icons/16x16/status/image-missing.png: Unrecognized image file format (gdk-pixbuf-error-quark, 3)
Bail out! Gtk:ERROR:gtkiconhelper.c:494:ensure_surface_for_gicon: assertion failed (error == NULL): Failed to load /org/gtk/libgtk/icons/16x16/status/image-missing.png: Unrecognized image file format (gdk-pixbuf-error-quark, 3)
Redirecting call to abort() to mozalloc_abort
```

So, something is wrong with this mime types database, in theory it should describe PNG. I've tried manually:
```
$ XDG_DATA_DIRS=/usr/share/qubes/mime xdg-mime query filetype test.png
application/x-zerosize
$ stat -c %s test.png
565
```
?!

Some earlier errors:
```
(thunderbird:2039): Gtk-WARNING **: 10:07:20.409: Could not load a pixbuf from /org/gtk/libgtk/theme/Adwaita/assets/bullet-symbolic.svg.
This may indicate that pixbuf loaders or the mime database could not be found.

(thunderbird:2039): Gtk-WARNING **: 10:07:21.352: Could not find the icon 'dialog-question'. The 'hicolor' theme
was not found either, perhaps you need to install it.
You can get a copy from:
	http://icon-theme.freedesktop.org/releases

(thunderbird:2039): Gtk-WARNING **: 10:07:21.352: Error loading theme icon 'dialog-question' for stock: Icon 'dialog-question' not present in theme Adwaita

(thunderbird:2039): Gtk-WARNING **: 10:07:21.352: Error loading theme icon 'image-missing' for stock: Failed to load /org/gtk/libgtk/icons/48x48/status/image-missing.png: Unrecognized image file format
```

The earlier messages suggests my concern about XDG_DATA_DIRS was correct:
> In fact, similar risk applies to $XDG_DATA_DIRS, for example /usr/share/icons, or /usr/share/fonts (not sure if the latter is sensitive to $XDG_DATA_DIRS). For this, probably some symlinks will be enough.

Indeed when making `/usr/share/qubes/mime/icons` a symlink to `/usr/share/icons`, I get correct icon paths, but it still complains about unknown image format.

marmarek:
> $ XDG_DATA_DIRS=/usr/share/qubes/mime xdg-mime query filetype test.png

It should be `/usr/share/qubes`. Then I get correctly `image/png`, and with this change applied to `qubes.StartApp` script too, Thunderbird doesn't crash anymore. But still doesn't want to open the file in DisposableVM by default (both other points stands).

DemiMarie:
> > $ XDG_DATA_DIRS=/usr/share/qubes/mime xdg-mime query filetype test.png
> 
> It should be `/usr/share/qubes`. Then I get correctly `image/png`, and with this change applied to `qubes.StartApp` script too, Thunderbird doesn't crash anymore. But still doesn't want to open the file in DisposableVM by default (both other points stands).

I diffed my working `/usr/local/share` with the non-working `/usr/share/qubes`, and it appears the problem is that `/usr/share/qubes/applications` is missing.

marmarek:
Ok, this improves the situation, but there is still a change that save is default instead of open. Is it intentional?

misc/applications/qvm-view-in-dvm.desktop
```diff
@@ -0,0 +1,7 @@
+[Desktop Entry]
+Version=1.0
+Name=QubesOS View In DisposableVM
+Exec=/usr/bin/qvm-open-in-dvm --view-only -- %U
```
marmarek, line 4: Should be `%u` - this tool does not support multiple files at once.

DemiMarie:
> Ok, this improves the situation, but there is still a change that save is default instead of open. Is it intentional?

No, it is not.

misc/Makefile
```diff
@@ -19,6 +20,11 @@ install:
 	    marker-vm \
 	    qubes-master-key.asc
 	install -D -m 0644 mime-globs $(DESTDIR)$(QUBESDATADIR)/mime-override/globs
+	ln -sf ../GConf $(DESTDIR)$(QUBESDATADIR)/GConf
+	ln -sf ../icons $(DESTDIR)$(QUBESDATADIR)/icons
+	ln -sf ../glib-2.0 $(DESTDIR)$(QUBESDATADIR)/glib-2.0
+	ln -sf ../../../etc/qubes/applications $(DESTDIR)$(QUBESDATADIR)/applications
```
marmarek, line 26: Hm? Shouldn't that be `cp -r applications $(DESTDIR)$(QUBESDATADIR)/` instead?

misc/Makefile
```diff
@@ -5,20 +5,28 @@ UDEVRULESDIR = $(SYSLIBDIR)/udev/rules.d
 QUBESDATADIR = $(DATADIR)/qubes
 
 .PHONY: all clean install
+.ONESHELL:
 
 all: marker-vm
+	export XDG_DATA_HOME=.; update-desktop-database . && exec update-mime-database mime
```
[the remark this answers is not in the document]
marmarek: And also, `update-mime-database mime` produces empty `mime/icons` that is not expected in the spec file, nor debian's install files. See all the build failures in the CI. I'm not sure if the file is actually needed - either remove it, or add to the packaging.

marmarek (changes requested):

marmarek:
I've made an automated tests for this feature: https://github.com/QubesOS/qubes-core-admin/pull/419
For now just Thunderbird, but adding other applications should be easy-ish. It probably doesn't work yet, but allowed me to detect some of the above listed issues.

DemiMarie (commented):
There is no mimeapps.list file, so everything defaults to “save” in Thunderbird.

misc/Makefile
```diff
@@ -5,20 +5,28 @@ UDEVRULESDIR = $(SYSLIBDIR)/udev/rules.d
...
 
 clean:
 	-$(RM) marker-vm
 
 install:
+	umask 022
```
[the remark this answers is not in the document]
marmarek: I disagree. `umask` relies on this being in the same target, while `install -m` is immune to moving around. And much easier to reason what is installed with what permissions.

Furthermore, `.ONESHELL` leads to surprising effects, as commented above.

marmarek:
> There is no mimeapps.list file, so everything defaults to “save” in Thunderbird.

With `update-desktop-database` called on the right dir (as commented above), `mimeapps.list` is correctly created and default action is "open" again.

marmarek (changes requested):

qubes-rpc/qubes.GetAppmenus
```diff
@@ -46,18 +46,17 @@ for dir in "${apps_dirs_to_consider[@]}"; do
...
-         BEGINFILE { if (ERRNO) nextfile; entry="" }
-         /^\[/ { if (tolower($0) != "\[desktop entry\]") nextfile } 
-         /^Exec *=/ { entry = entry FILENAME ":Exec=qubes-desktop-run " FILENAME "\n"; next }
-         /^NoDisplay *= *true$/ { entry=""; nextfile }
-         /=/ { entry = entry FILENAME ":" $0 "\n" }
-         ENDFILE { print entry }
-         ' 2> /dev/null
+for i in "${apps_dirs[@]}"; do
+    case $i in (/*) :;; (*) continue;; esac
+    (cd "$i" && exec find . -name '*.desktop' -print0 2>/dev/null)
```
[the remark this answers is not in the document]
DemiMarie: That is a great idea, thanks!

andrewdavidwong:
The issue https://github.com/QubesOS/qubes-issues/issues/6877 has been created for this.

marmarek (approved):

PR #292, opened 2021-01-22
Avoid passing options to ‘dnf clean’

DemiMarie:
They are useless and can cause errors.


package-managers/qubes-download-dom0-updates.sh
```diff
@@ -84,7 +84,7 @@ rpm --root=$DOM0_UPDATES_DIR --rebuilddb
 
 if [ "$CLEAN" = "1" ]; then
     # shellcheck disable=SC2086
-    "${YUM[@]}" "${OPTS[@]}" clean all
+    "${YUM[@]}" clean all
```
marmarek, line 87: This won't work this easily. `$OPTS` contains `--installroot` etc, without them it will clean the wrong cache.

marmarek (changes requested):

marmarek:
@DemiMarie can you fix this? It prevents merging some of the changes in https://github.com/QubesOS/qubes-mgmt-salt-dom0-update/pull/14 (when we get to the working solution).

marmarek (approved):

PR #303, opened 2021-05-14
Export DNS information obtained during firewall setup to QubesDB

3hhh:
This can e.g. allow DNS applications to pin a FQDN to the IP
used during Qubes OS firewall setup.

Can help with QubesOS/qubes-issues#5225 and other related issues.

### DIFF
--- a/qubesagent/firewall.py
+++ b/qubesagent/firewall.py
@@ -127,6 +127,26 @@ class FirewallWorker(object):
         rules.append({'action': policy})
         return rules
 
+    def resolve_dns(self, fqdn, family):
+        """
+        Resolve the given FQDN via DNS.
+        :param fqdn: FQDN
+        :param family: 4 or 6 for IPv4 or IPv6
+        :return: see socket.getaddrinfo()
+        :raises: RuleParseError
+        """
+        try:
+            addrinfo = socket.getaddrinfo(fqdn, None,
+                (socket.AF_INET6 if family == 6 else socket.AF_INET))
+        except socket.gaierror as e:
+            raise RuleParseError('Failed to resolve {}: {}'.format(
+                fqdn, str(e)))
+        except UnicodeError as e:
+            raise RuleParseError('Invalid destination {}: {}'.format(
+                fqdn, str(e)))
+        return addrinfo
+
+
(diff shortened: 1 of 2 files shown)


3hhh:
Wrt failing tests: I guess I'll fix them if the general approach is OK for upstream. I tested the code manually though and it worked.

marmarek:
Generally I think this approach is fine. But specifically, I'd propose few improvements:
1. keep `prepare_rules` as pure function (without side effects), return resolved addresses (like, make it return a tuple of prepared rules + addresses?), and write them into qubesdb only when actual rules gets applied - this way you won't (over)write entries if rules won't get applied (because of some error later for example)
2. consider adding some hook script to be called after updating firewall + writing those addresses, so you can integrate it easier with whatever caching dns resolver you want; in fact, maybe restore calling firewall-user-script after each firewall change? it was this way in Qubes 3.2; maybe also add some parameter (or env variable) with info for what vm (address) rules were reloaded

3hhh:
Thanks for your review!

I implemented your points (2. via QubesDB though), but still need to fix the tests / some introduced bugs and possibly add some checks for the new features in the tests.

Unrelated potential race condition bugs I found during my review & tests:
- https://github.com/QubesOS/qubes-core-agent-linux/blob/ed33374f6722ea4fc7fa967df111da6f8b507d8d/qubesagent/firewall.py#L184-L190
  I'd have expected the watches first before setting the initial state. If there's an update right after the init, the watch will miss it as it isn't watching yet. But then I didn't look at the qubesdb code very much.
- VMs are never removed from iptables/nft rules & QubesDB. This should be fine as long as an IP is never assigned twice during a Qubes session (I don't know atm). If it is, there may be a short period of time where the new VM will have the firewall rules of the old VM.

3hhh:
That should be it then.

I also added a fix for the aforementioned presumed first bug: Just cherry-pick https://github.com/QubesOS/qubes-core-agent-linux/pull/303/commits/795bec8038c948fd0179452ecb70f7eccea472a7 away if you deem it incorrect.

marmarek (changes requested):

marmarek (approved):

PR #304, opened 2021-05-14
fix #6562. qvm-open-in-vm: wait EOF when handling URL

neowutran:
should fix QubesOS/qubes-issues#6562.
The assumption that I made is that there is no case where we need to have a non-empty stdout when handling URL in qvm-open-in-vm.

(At least, tested if against my firefox issue, and it work as expected. But didn't ran more extensive tests)


marmarek (approved):

StylusAverageTremor:
Is this going to be backported to 4.0?

PR #305, opened 2021-05-15
Include official Qubes repositories

unman:
Addresses QubesOS/qubes-issues#6610 by setting official repository definitions (at least for 4.0 - I'll check 4.1 later)


marmarek:
It's in https://github.com/QubesOS/qubes-meta-packages/

unman:
On Sat, May 15, 2021 at 03:25:38PM -0700, Marek Marczykowski-G??recki wrote:
> It's in https://github.com/QubesOS/qubes-meta-packages/
> 

So it is, but the package isnt installed in the template, so that
package is pointless at the moment.
Is this deliberate? An oversight?

PR #314, opened 2021-06-21
firewall: make notify-send work

3hhh:
Manually tested in debian-10 and fedora-33.

Fixes QubesOS/qubes-issues#6658

### DIFF
--- a/qubesagent/firewall.py
+++ b/qubesagent/firewall.py
@@ -24,6 +24,7 @@ import logging
 import os
 import socket
 import subprocess
+import pwd
 import shutil
 import daemon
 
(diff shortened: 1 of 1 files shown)


qubesagent/firewall.py
```diff
@@ -185,10 +185,20 @@ def is_ip6(addr):
 
     def log_error(self, msg):
         self.log.error(msg)
-        subprocess.call(
-            ['notify-send', '-t', '3000', msg],
-            env=os.environ.copy().update({'DISPLAY': ':0'})
-        )
+        try:
+            subprocess.check_output(
+                ['sudo', '-u', self.qdb.read('/default-user') or 'user',
```
marmarek, line 190: Maybe use `runuser` instead, to avoid dependence on sudo config (and sudo being installed at all)?

marmarek (changes requested):

qubesagent/firewall.py
3hhh: I did that. However `runuser` unfortunately requires `DBUS_SESSION_BUS_ADDRESS` on fedora-33, which makes the code pretty ugly (`sudo` seems to do more environment var magic?). Cherry-pick whatever variant you prefer.

It might be worth encapsulating notification code in some Qubes class with screenshot testing etc. long term.

marmarek (approved):

PR #321, opened 2021-08-06
qubes.Syslog

3hhh:
This is a qrexec service to receive logs.

Fixes QubesOS/qubes-issues#830

All the doc is in the file header.

The packaging should work. I don't see the point of providing a default RPC policy for this qrexec service as it requires user intervention anyway.

### DIFF
--- a/qubes-rpc/qubes.Syslog
+++ b/qubes-rpc/qubes.Syslog
@@ -1,0 +1,152 @@
+#!/usr/bin/env python3
+# vim: fileencoding=utf-8
+#
+# The Qubes OS Project, https://www.qubes-os.org/
+#
+# Copyright (C) 2021
+#                   David Hobach <tripleh@hackingthe.net>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+'''
+Qrexec service to forward logs to. Useful to forward logs from multiple VMs to
+a single VM.
+
+All logs sent to this service are roughly sanitized and then passed over to the
+configured system logger (usually systemd/journald listening at `/dev/log`
+nowadays).
+
+A single log entry must be UTF-8 encoded and fit on a single line ending with `\n`.
+
+Logs may have any of the following two formats:
+1. [message]       (example: `this is a log entry`)
+2. [PRI] [message] (example: `<165>This is another log entry.`)
+
+[PRI]:     As defined in [RFC3164] and [RFC5424].
+[message]: An arbitrary string without newline. This service removes
+           non-printable characters and truncates messages at 1500 characters.
+
+Please note that the internal message structure is not sanitized. So any receiving
+application should perform its own sanitization, if the message is structured
+further.
+
+For example, a user could deploy the following rsyslog configuration in a sending
+VM at `/etc/rsyslog.d/qubes_syslog.conf` (requires `systemctl restart rsyslog`):
+```
+module(load="omprog")
+
+template(name="qubes_syslog" type="string"
+    string="<%PRI%>%TIMESTAMP:::date-rfc3339% %syslogtag%%msg:::sp-if-no-1st-sp%%msg:::drop-last-lf%\n")
+
+ruleset(name="qubes") {
+  action(
+    type="omprog"
+    binary="/usr/sbin/runuser -u user -- /usr/bin/qrexec-client-vm work qubes.Syslog"
+    template="qubes_syslog"
+  )
+}
+
+#apply your filtering here (*.* just forwards all logs)
+*.* call qubes
+```
+
+Users can also e.g. pipe stored log files to a VM via
+`qrexec-client-vm vm qubes.Syslog < /path/to/some/logfile`.
+
+References:
+[RFC3164]: https://datatracker.ietf.org/doc/html/rfc3164
+[RFC5424]: https://datatracker.ietf.org/doc/html/rfc5424
+'''
+
+import io
+import sys
+import os
+import logging
+import syslog
+import re
+import unicodedata
+
+def get_logger():
+    log = logging.getLogger("qubes.Syslog")
+    log.setLevel(logging.INFO)
+    return log
+
+#globals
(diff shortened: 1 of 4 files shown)


3hhh:
KISS ftw!

qubes-rpc/qubes.Syslog
```diff
@@ -0,0 +1,152 @@
...
+    return SEV2SYSLOG[severity]
+
+def filter_non_printable(string):
+    ret = []
+    for ch in string:
+        cat = unicodedata.category(ch)
+        if cat[0] == "C":
+            ret += "_"
+        else:
+            ret += ch
```
marmarek, line 111: I'd be more strict about it: just printable ASCII only. Rendering unicode is a complex thing, especially as you can combine various characters etc. Lets not expose whatever you use for viewing the logs...

marmarek (changes requested):

qubes-rpc/qubes.Syslog
3hhh: I don't like to exclude all non-American non-European logs.

What a receiving application should do: Don't assume anything about the encoding, consider it a byte stream and store it somewhere. Whatever viewer then will attempt to render that byte stream (probably `less`, `vim` or whatever pager `journalctl` uses). [rsyslog](https://www.rsyslog.com/doc/v7-stable/whitepapers/syslog_protocol.html) seems to do it that way.

[RFC5424](https://datatracker.ietf.org/doc/html/rfc5424) even clearly states:
"The character set used in MSG SHOULD be UNICODE, encoded using UTF-8
 as specified in [RFC3629].  If the syslog application cannot encode
 the MSG in Unicode, it MAY use any other encoding." [RFC3164](https://datatracker.ietf.org/doc/html/rfc3164) doesn't have such encoding statements which probably resulted in the aforementioned rsyslog finding ("I have now simply taken the message as is and stuffed it into the MSG part").

I also think that users can understand the implications of working with potentially untrustworthy logs.

Btw if such a bug via non-ASCII characters existed, all Linux distributions would be affected by privilege escalation whenever a root user would view system logs via e.g. `journalctl -b0`. A user would just have to do something like `logger "exploit here"`... Not impossible, but rather unlikely from my point of view.

fepitre: I would say that for a first version and until we are sure there is no security implication or side effect, I would suggest to restrict to what @marmarek said.

marmarek: I understand this limitation would affect some people more than others, but also, the main focus of Qubes is to provide reasonably secure system, and rendering unicode does have significant attack surface. This is not purely theoretical, see CVE-2019-18397 for example or almost yearly incident of an unicode wifi name or message crashing iPhones.
Excluding control characters reduces that risk significantly, but IMO not enough to have it allowed by default. That's also why we are not allowing unicode characters in window titles by default.

> I also think that users can understand the implications of working with potentially untrustworthy logs.

No, I don't think so. Many (most?) can't understand that doing "curl | bash" is a horrible idea, or pasting something copied from a website directly to a shell (even if the command looks harmless: https://thejh.net/misc/website-terminal-copy-paste). So I definitely wouldn't expect people to understand that reading logs (which is seemingly much safer operation) may be risky if they come from untrusted place. When reading logs in a logvm, it may even not be obvious that they do come from untrusted place in the first place.

> Btw if such a bug via non-ASCII characters existed, all Linux distributions would be affected by privilege escalation whenever a root user would view system logs via e.g. journalctl -b0. A user would just have to do something like logger "exploit here"... Not impossible, but rather unlikely from my point of view.

Here you go: https://seclists.org/fulldisclosure/2021/May/51
This one relies on control characters that you excluded, but just a proof that printing untrusted data in a terminal is still (in 2021) not safe yet.

3hhh: OK, that proves you right.

I made it ASCII by default now, but provided an option for unicode support.

If that approach is acceptable, I'd adjust my other PR accordingly.

Btw wrt security considerations:
The logs are written via the libc `syslog` routine, which then writes to the `/dev/log` unix socket, which usually `journald` listens to. `rsyslog` - if installed & configured - then obtains the logs via its `imjournal` module from `journald` and does whatever it's configured to do with them. Some distros (e.g. debian) don't use `imjournal`, but [imuxsock](https://www.rsyslog.com/doc/v8-stable/configuration/modules/imuxsock.html#imuxsock-systemd-details-label) instead.

qubes-rpc/qubes.Syslog
```diff
@@ -0,0 +1,164 @@
...
+1. [message]       (example: `this is a log entry`)
+2. [PRI] [message] (example: `<165>This is another log entry.`)
+
+[PRI]:     As defined in [RFC3164] and [RFC5424].
+[message]: An arbitrary string without newline. This service removes
+           non-printable characters and truncates messages at 1500 characters.
+           By default non-ASCII characters are replaced with question marks.
+           To allow unicode, set QubesDB `/qrexec/qubes-syslog/allow-unicode` to
+           `True`. This infers security risks with unicode parsing bugs in log
+           viewers.
```
DemiMarie, line 44: Another option would be to losslessly escape them, so that no information is lost.

qubes-rpc/qubes.Syslog
```diff
@@ -0,0 +1,164 @@
...
+        untr = untr.encode(encoding="ascii", errors="replace").decode("ascii")
+
+    return prio, filter_non_printable(untr)
+
+def handle_untrusted(vm, allow_unicode=False):
+    pid = os.getpid()
+    syslog.openlog(ident=f"qubes.Syslog|{vm}[{pid}]", facility=syslog.LOG_USER)
+
+    input_stream = io.TextIOWrapper(sys.stdin.buffer, encoding="utf-8")
+    for untrusted_line in input_stream:
```
DemiMarie, line 141: This allows a trivial DoS by sending a very very very long line.

qubes-rpc/qubes.Syslog
3hhh: Yes, a sender may e.g. encode all logs in base64 and then decode it in his super trusted log viewer. ^^

Until such a log viewer exists however I'll assume that automatic escaping/unescaping only provides more bug opportunities in realistic viewers.

Btw a Qubes OS-like solution to this viewer distrust would be to accept unicode and then allow viewing the logs only in a dispVM.

qubes-rpc/qubes.Syslog
3hhh: Yup. I didn't bother about DoS as there's many opportunities in Qubes to do something like that (e.g. use the CPU a lot, I/O a lot, send tons of useless logs making journald rotate out the relevant ones, ...). Since it's easy to fix, I fixed it though.

ddevz:
Hey, I just wanted to say that this is very exciting and thank you for doing this!

qubes-rpc/qubes.Syslog
DemiMarie: > Btw a Qubes OS-like solution to this viewer distrust would be to accept unicode and then allow viewing the logs only in a dispVM.

I love this solution!

3hhh:
@marmarek Does this PR still have a chance to go upstream or should I close it down?

PR #322, opened 2021-08-23
Require ‘locales-all’ package

DemiMarie:
Installing only the `locales` package doesn’t install support for `en_US.UTF-8`, which causes breakage.

Fixes QubesOS/qubes-issues#4889.


marmarek:
No, `locales-all` definitely should not be a hard dependency here. It takes about 230MB, which means +30% for minimal template size.
Package description:
> This package contains the precompiled locale data for all supported locales. A better alternative is to install the locales package and only select desired locales, but it can be useful on a low-memory machine because some locale files take a lot of memory to be compiled. 

So, it sounds like installing this packages is not a good idea at all. Instead, appropriate locale should be generated.

PR #324, opened 2021-08-27
Set the SELinux context of xendriverdomain.service

DemiMarie:
This is a step towards having SELinux enforcing in Fedora qubes.


marmarek:
Already merged in another PR (#343)

PR #342, opened 2021-11-29
fix upgrades-installed-check for archlinux

kevinjohna6:
set -e causes an early exit rather than correct behavior when checkupdates returns 2.
This PR makes it so that the script only prints true or false when exit codes are valid, and otherwise exits with the error code.

For context, I needed to fix this to pass qubes-dist-upgrade stage 1 normally.


marmarek (approved):
Approved as of 113cb07d3c9463af0b77ee4b245d6a5982336fb3

PR #344, opened 2021-12-13
qubes-vm-core: allow python<3.11 as a dependency

seberm:
The python v3.10 is finally out in the official arch repositories:
- https://archlinux.org/packages/core/x86_64/python/

The python dependecy should be updated. This change should be also backported to PKGBUILD for Qubes v4.0.

@lubellier, could you help me to test this change?

I have tested the change in my arch template and it seems that everything works as expected.

Thanks,
O.

### DIFF
--- a/archlinux/PKGBUILD
+++ b/archlinux/PKGBUILD
@@ -67,7 +67,7 @@ package_qubes-vm-core() {
              python-dbus xdg-utils notification-daemon gawk sed procps-ng librsvg
              socat pacman-contrib
              # Block updating if there is a major python update as the python API will be in the wrong PYTHONPATH
-             'python<3.10'
+             'python<3.11'
              )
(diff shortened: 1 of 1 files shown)


lubellier:
I will test it tomorrow.
[qubes-db-vm PKGBUILD](https://github.com/QubesOS/qubes-core-qubesdb/blob/master/archlinux/PKGBUILD) should also be updated.

marmarek (approved):
Approved as of da8fa95bfc89846788c582c11c5afa51ae478bc2

james-callahan:
> This change should be also backported to PKGBUILD for Qubes v4.0.

Did this backport happen?

lubellier:
> Did this backport happen?

No, look [PKGBUILD#L66](https://github.com/QubesOS/qubes-core-agent-linux/blob/release4.0/archlinux/PKGBUILD#L66) in the 4.0-branch, it's `python<3.10` not `python<3.11`.

PR #346, opened 2021-12-16
Fail boot if random seed doesn't get initialized

DemiMarie:
The random seed seems to (rarely) not get initialized.  This will ensure
that the random seed must be initialized for the system to successfully
boot.


DemiMarie:
PipelineRetry

DemiMarie:
PipelineRetry

marmarek (approved):

PR #347, opened 2021-12-16
Support nanosecond precision for dates

DemiMarie:
Requires QubesOS/qubes-core-admin#442 to be installed in dom0.
Fixes QubesOS/qubes-issues#7081.

### DIFF
--- a/qubes-rpc/qubes-sync-clock
+++ b/qubes-rpc/qubes-sync-clock
@@ -53,9 +53,14 @@ def main():
         sys.stderr.write('Received a date older than this program, aborting!\n')
         sys.exit(1)
     date_out = untrusted_date_out
-    subprocess.check_call([b'date', b'-u', b'-I' + precision, b'-s',
-                           date_out[:-1]],
-        stdout=subprocess.DEVNULL)
+    try:
+        subprocess.check_call([b'date', b'-u', b'-I' + precision, b'-s',
+                               date_out[:-1]],
+            stdout=subprocess.DEVNULL)
+    except subprocess.CalledProcessError as e:
+        # input is trusted here, so it can be safely printed
+        sys.stderr.write('Unable to set the date: process {!r} failed.\n'.format(e.cmd))
+        sys.exit(e.returncode)
 
 if __name__ == '__main__':
     main()
--- a/qubes-rpc/qubes-sync-clock
+++ b/qubes-rpc/qubes-sync-clock
@@ -18,20 +18,43 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #
 
-import sys
+import os
 import re
 import subprocess
+import sys
 
 def main():
-    stdin = sys.stdin.read(25)
-
-    untrusted_date_out = stdin.strip()
-
-    if not re.match(r'^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+00:?00$', untrusted_date_out):
+    os.environ['LC_ALL'] = 'C'
+    untrusted_date_out = sys.stdin.buffer.read(36)
+    try:
+        untrusted_date_out.decode('ascii', 'strict')
+    except UnicodeDecodeError:
+        sys.stderr.write('Received non-ASCII date, aborting!\n')
+        sys.exit(1)
+    untrusted_date_len = len(untrusted_date_out)
+    if untrusted_date_len == 36: # new format, nanosecond precision
+        regexp = rb'\A\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2},[0-9]{9}\+00:00\n\Z'
+        precision = b'ns'
+    elif untrusted_date_len == 26: # old format, second precision
+        regexp = rb'\A\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+00:00\n\Z'
+        precision = b'seconds'
+    else:
+        sys.stderr.write('Invalid date length (expected 26 or 36 bytes, got {})'
+                         ', aborting!\n'.format(untrusted_date_len))
+        sys.exit(1)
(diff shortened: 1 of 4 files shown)


marmarek:
To make upgrade path less risky, I prefer to keep output format of the current service unchanged, and introduce new service with ns precision instead.

Then, you can also announce support for the new one via qvm-features (see `qubes-rpc/post-install.d` dir in this repo). Currently we don't have generic "is qrexec service X supported" namespace there, but maybe we should have it? An alternative is to always try `qubes.GetDateNS` first and fallback to `qubes.GetDate` if it fails with 127 exit code (we do this in few places already). Not a bad alternative in a long run, but for the migration period it may be annoying.

DemiMarie:
> To make upgrade path less risky, I prefer to keep output format of the current service unchanged, and introduce new service with ns precision instead.

What about using the service argument for this?  The current dom0 code will pass an empty string, and the current VM-side code will ignore the option.

> Currently we don't have generic "is qrexec service X supported" namespace there, but maybe we should have it?

This is a good idea, but using it is more complicated than it needs to be here.

(What I really would like to do is to use a fixed-size binary or hex number.  That makes backwards compatibility much more difficult, though.)

marmarek:
> What about using the service argument for this? The current dom0 code will pass an empty string, and the current VM-side code will ignore the option.

That would work too.

qubes-rpc/qubes.GetDate
```diff
@@ -16,7 +16,10 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
-#
-#
 
-date -u -Iseconds
+export LC_ALL=C
+if [ -n "${1-}" ] && [ "$1" -ge 1 ]; then
```
marmarek, line 21: Maybe make the argument a bit more descriptive? Like "nanoseconds", or "ns"?

marmarek:
And also, the format should explicitly be documented at https://www.qubes-os.org/doc/vm-interface/#qubes-rpc

qubes-rpc/qubes.GetDate
DemiMarie: The argument was intended to be a version number, in case we decide to change this again in the future.  I can use a more descriptive string instead, though.

DemiMarie:
> And also, the format should explicitly be documented at https://www.qubes-os.org/doc/vm-interface/#qubes-rpc

Will do (in a separate PR)

qubes-rpc/qubes-sync-clock
```diff
@@ -23,16 +23,42 @@ import re
 import subprocess
 
 def main():
-    stdin = sys.stdin.read(25)
-
-    untrusted_date_out = stdin.strip()
-
-    if not re.match(r'^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+00:?00$', untrusted_date_out):
+    untrusted_date_out = sys.stdin.read(36)
```
marmarek, line 26: `sys.stdin.read()` returns `str`, not `bytes`. All the below handling of `bytes` (regex, comparing etc) are wrong.
If you really want bytes, then use `sys.stdin.buffer.read()`, but even then, save result of the decode call below and operate on str further down.

marmarek (changes requested):

qubes-rpc/qubes-sync-clock
DemiMarie: I will fix QubesOS/qubes-core-admin#442 and then copy the changed code to here.

DemiMarie:
PipelineRetry

marmarek:
From openQA run:
```
[2022-07-09 13:51:08] [   18.536237] qvm-sync-clock[638]: Traceback (most recent call last):
[2022-07-09 13:51:08] [   18.536944] qvm-sync-clock[638]:   File "/usr/lib/qubes/qubes-sync-clock", line 65, in <module>
[2022-07-09 13:51:08] [   18.537634] qvm-sync-clock[638]:     main()
[2022-07-09 13:51:08] [   18.537919] qvm-sync-clock[638]:   File "/usr/lib/qubes/qubes-sync-clock", line 26, in main
[2022-07-09 13:51:08] [   18.539561] qvm-sync-clock[638]:     os.environ['LC_ALL'] = 'C'
[2022-07-09 13:51:08] [   18.540891] qvm-sync-clock[638]: NameError: name 'os' is not defined
[2022-07-09 13:51:08] [   18.575022] systemd[1]: qubes-sync-time.service: Succeeded.
```
Besides the obvious error, I wonder why systemd consider service as successful...

DemiMarie:
> From openQA run:
> 
> ```
> [2022-07-09 13:51:08] [   18.536237] qvm-sync-clock[638]: Traceback (most recent call last):
> [2022-07-09 13:51:08] [   18.536944] qvm-sync-clock[638]:   File "/usr/lib/qubes/qubes-sync-clock", line 65, in <module>
> [2022-07-09 13:51:08] [   18.537634] qvm-sync-clock[638]:     main()
> [2022-07-09 13:51:08] [   18.537919] qvm-sync-clock[638]:   File "/usr/lib/qubes/qubes-sync-clock", line 26, in main
> [2022-07-09 13:51:08] [   18.539561] qvm-sync-clock[638]:     os.environ['LC_ALL'] = 'C'
> [2022-07-09 13:51:08] [   18.540891] qvm-sync-clock[638]: NameError: name 'os' is not defined
> [2022-07-09 13:51:08] [   18.575022] systemd[1]: qubes-sync-time.service: Succeeded.
> ```

Python code fixed.  Tested by checking that I get the expected “Operation not permitted error” when not running as root on correct input.

> Besides the obvious error, I wonder why systemd consider service as successful...

Fixed by using `Type=oneshot`.

marmarek (approved):

PR #351, opened 2021-12-21
Enable proxy NDP if IPv6 is enabled

DemiMarie:
This is consistent with proxy ARP; if one is enabled, the other should be as well.

The first commit improves error handling for IPv6 being disabled.  The second commit is the actual fix, and the third commit removes an unused variable.

### DIFF
--- a/network/vif-route-qubes
+++ b/network/vif-route-qubes
@@ -102,7 +102,7 @@ if ! [[ $vif =~ ^vif(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)$ ]]; then
     exit 1
 fi
 
-domid=${BASH_REMATCH[1]} sub=${BASH_REMATCH[2]}
+domid=${BASH_REMATCH[1]}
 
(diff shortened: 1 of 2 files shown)


marmarek:
shellcheck complains

network/vif-route-qubes
```diff
@@ -120,9 +120,7 @@ case "$command" in
         ipcmd='add'
         iptables_cmd='-I PREROUTING 1'
         cmdprefix=''
-        ipv6_disabled=$(cat /proc/sys/net/ipv6/conf/"${vif}"/disable_ipv6 || echo 1)
-        # without a MAC address we will fail later with a confusing error
-        mac=$(xenstore-read "backend/vif/$domid/$sub/mac") || exit 1
```
marmarek, line 125: It is still used in `vif-qubes-nat.sh` script. I think this line is responsible for all fake_ip tests failures.

marmarek (approved):

PR #352, opened 2021-12-25
dnf: drop old dnf support

fepitre:
Fixes https://github.com/QubesOS/qubes-issues/issues/7083

### DIFF
--- a/package-managers/dnf-qubes-hooks.py
+++ b/package-managers/dnf-qubes-hooks.py
@@ -20,10 +20,9 @@
 #
 
 from __future__ import absolute_import
-from distutils.version import LooseVersion
+
 import logging
 import dnf
-import dnf.const
 import subprocess
 
 PLUGIN_CONF = 'qubes-hooks'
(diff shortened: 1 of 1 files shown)


fepitre:
PipelineRetry

package-managers/dnf-qubes-hooks.py
```diff
@@ -20,9 +20,8 @@
 #
 
 from __future__ import absolute_import
-from distutils.version import LooseVersion
+
 import logging
-import dnf
```
marmarek, line 25: I think you meant removing `import dnf.const` instead. `import dnf` is used for `dnf.Plugin` base class.

marmarek (changes requested):

package-managers/dnf-qubes-hooks.py
fepitre: Yes.

marmarek (approved):

PR #354, opened 2022-01-07
Harden network interfaces before enabling them

DemiMarie:
This is done via various sysctls.

### DIFF
--- a/network/80-qubes.conf
+++ b/network/80-qubes.conf
@@ -1,18 +1,1 @@
-# Must be set system-wide
-net.ipv6.max_hbh_opts_length = 0
-net.ipv6.max_dst_opts_length = 0
-net.ipv6.conf.all.accept_dad = 0
-net.ipv4.conf.all.accept_redirects = 0
-
-net.ipv4.conf.*.accept_source_route = 0
-net.ipv4.conf.*.accept_redirects = 0
-net.ipv4.conf.*.secure_redirects = 0
-net.ipv4.conf.*.send_redirects = 0
-net.ipv4.conf.*.drop_unicast_in_l2_multicast = 1
-net.ipv6.conf.*.accept_source_route = -1
-net.ipv6.conf.*.accept_redirects = 0
-net.ipv6.conf.*.accept_ra = 0
-net.ipv6.conf.*.accept_dad = 0
-net.ipv6.conf.*.autoconf = 0
-net.ipv6.conf.*.drop_unicast_in_l2_multicast = 1
 net.ipv6.conf.*.drop_unsolicited_na = 1
--- a/network/80-qubes.conf
+++ b/network/80-qubes.conf
@@ -1,1 +1,18 @@
-net.ipv6.conf.all.drop_unsolicited_na=1
+# Must be set system-wide
+net.ipv6.max_hbh_opts_length = 0
+net.ipv6.max_dst_opts_length = 0
+net.ipv6.conf.all.accept_dad = 0
+net.ipv4.conf.all.accept_redirects = 0
+
+net.ipv4.conf.*.accept_source_route = 0
+net.ipv4.conf.*.accept_redirects = 0
+net.ipv4.conf.*.secure_redirects = 0
+net.ipv4.conf.*.send_redirects = 0
+net.ipv4.conf.*.drop_unicast_in_l2_multicast = 1
+net.ipv6.conf.*.accept_source_route = -1
+net.ipv6.conf.*.accept_redirects = 0
+net.ipv6.conf.*.accept_ra = 0
+net.ipv6.conf.*.accept_dad = 0
+net.ipv6.conf.*.autoconf = 0
+net.ipv6.conf.*.drop_unicast_in_l2_multicast = 1
+net.ipv6.conf.*.drop_unsolicited_na = 1
--- a/network/setup-ip
+++ b/network/setup-ip
@@ -24,22 +24,6 @@ configure_network() {
     local gateway6="$8"
     local primary_dns="$9"
     local secondary_dns="${10}"
-    local i
-
-    for i in all "$INTERFACE"; do
-        echo 0 > "/proc/sys/net/ipv4/conf/$i/accept_source_route"
-        echo 0 > "/proc/sys/net/ipv4/conf/$i/accept_redirects"
-        echo 0 > "/proc/sys/net/ipv4/conf/$i/secure_redirects"
-        echo 0 > "/proc/sys/net/ipv4/conf/$i/send_redirects"
-        echo 1 > "/proc/sys/net/ipv4/conf/$i/drop_unicast_in_l2_multicast"
-        echo -1 > "/proc/sys/net/ipv6/conf/$i/accept_source_route"
-        echo 0 > "/proc/sys/net/ipv6/conf/$i/accept_redirects"
-        echo 0 > "/proc/sys/net/ipv6/conf/$i/accept_ra"
-        echo 0 > "/proc/sys/net/ipv6/conf/$i/accept_dad"
-        echo 0 > "/proc/sys/net/ipv6/conf/$i/autoconf"
-        echo 1 > "/proc/sys/net/ipv6/conf/$i/drop_unicast_in_l2_multicast"
-        echo 1 > "/proc/sys/net/ipv6/conf/$i/drop_unsolicited_na"
-    done
 
     /sbin/ip address add "$ip/$netmask" dev "$INTERFACE"
     if [ -n "$ip6" ]; then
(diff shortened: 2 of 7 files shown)


vm-systemd/qubes-network-uplink@.service
```diff
@@ -3,6 +3,8 @@ Description=Qubes network uplink (%i) setup
 After=network-pre.target qubes-iptables.service
 After=sys-subsystem-net-devices-%i.device
 BindsTo=sys-subsystem-net-devices-%i.device
+After=systemd-sysctl.service
+BindsTo=systemd-sysctl.service
```
marmarek, line 7: I don't think `BindsTo` is correct here, stopping interfaces together with systemd-sysctl doesn't make sense.

vm-systemd/qubes-network-uplink@.service
```diff
@@ -3,6 +3,8 @@ Description=Qubes network uplink (%i) setup
 After=network-pre.target qubes-iptables.service
 After=sys-subsystem-net-devices-%i.device
 BindsTo=sys-subsystem-net-devices-%i.device
+After=systemd-sysctl.service
```
marmarek, line 6: This is already implicitly here (`systemd-sysctl.service` has `Before=sysinint.target` and everything else has `After=sysinit.target` unless `DefaultDependencies=no`).

network/80-qubes.conf
```diff
@@ -1 +1,12 @@
...
+net.ipv4.conf.*.secure_redirects = 0
+net.ipv4.conf.*.send_redirects = 0
+net.ipv4.conf.*.drop_unicast_in_l2_multicast = 1
+net.ipv6.conf.*.accept_source_route = 0
+net.ipv6.conf.*.accept_redirects = 0
+net.ipv6.conf.*.accept_ra = 0
+net.ipv6.conf.*.accept_dad = 0
+net.ipv6.conf.*.autoconf = 0
+net.ipv6.conf.*.drop_unicast_in_l2_multicast = 1
+net.ipv6.conf.*.drop_unsolicited_na = 1
```
marmarek, line 12: This is rather massive number of changed settings, too big risk to get it into R4.1 at this point.

marmarek (changes requested):

network/80-qubes.conf
DemiMarie: Most of (all but `net.ipv[46].conf.*.drop_unicast_in_l2_multicast`) come from @fepitre’s [standard hardening configuration for CentOS][1], which in turn comes from the CIS CentOS 7 benchmark.  So I consider these to be very low-risk.  `net.ipv[46].conf.*.drop_unicast_in_l2_multicast` causes Linux to drop bogus packets that do not conform to standards, but it is the least important and I can get rid of it.

[1]: https://github.com/fepitre/centos-builder/blob/master/centos-hardened.ks#L192-L212

vm-systemd/qubes-network-uplink@.service
DemiMarie: :+1:

vm-systemd/qubes-network-uplink@.service
DemiMarie: I will use `Requires` instead.

DemiMarie: I wound up omitting it since all of the hardening is also done in the script itself.

network/80-qubes.conf
fepitre: It looks like this file is installed only on Debian system. I don't know why but for me it looks lie a duplicate with respect to `setup-ip` and `vif-route-qubes`.

network/setup-ip
```diff
@@ -24,6 +24,22 @@ configure_network() {
     local gateway6="$8"
     local primary_dns="$9"
     local secondary_dns="${10}"
+    local i
+
+    for i in all "$INTERFACE"; do
```
fepitre, line 29: Those options should be harmless but I'm not totally sure in the case where `$INTERFACE` serves a full separate network in the case of bridge setup. In all the case and in order to not "brick" any production system, I would propose to enable those sysctl options only if we request the specific qube to have the tag for example: `net.hardening`. Across all of our configurations, we may introduce hardening features as tag like `{net,system,etc.}.hardening`.

network/80-qubes.conf
marmarek: > However, I am not aware of any way vif-route-qubes could do that.

`vif-route-qubes` is called by `xendriverdomain` service, as part of bringing up `vif` interface. And `systemd-sysctl` is guaranteed to start earlier, since it starts before `sysinit.target`.

DemiMarie: To elaborate: I meant *for the specific network interface*.  It is trivial to set the `all` and `default` entries, but not the interface-specific ones.  If we had `Requires=systemd-sysctl.service` on `xendriverdomain.service`, that would likely be enough.  Should I do this with sysctl drop-ins, the script, or both?

network/setup-ip
DemiMarie: That is a good idea, especially if combined with the qubesdb fixes (https://github.com/QubesOS/qubes-core-qubesdb/pull/44).

fepitre: I also wonder if this should not apply to *only* "$INTERFACE" because *all* is already for all existing interfaces. I'm not sure how this would behave for example with docker, I would need to try.

network/80-qubes.conf
marmarek: >  If we had Requires=systemd-sysctl.service

That's not necessary. `systemd-sysctl.service` is started always (it is static unit). If you mean ordering, then that's common misconception about dependencies in systemd - `Requires` is about _if_ the unit will be started, not _when_.

marmarek: In any case, please decide on one solution - either sysctl, or scripts configuring specific interfaces, but not both. I'd prefer sysctl, because it's possible to override it for the user, if specific use case would require rolling back some of those settings.

network/setup-ip
DemiMarie: I believe this is sysctl-specific:

- `accept_seg6` and `rpl_seg_enabled` should both be set to 0 on the interface, and doing so is sufficient.
- `ipv6.conf.$i.accept_source_route` should be set to -1 (_not_ 0!), and doing so is sufficient.
- global sysctls `net.ipv6.max_hbh_opts_length` and `net.ipv6.max_dst_opts_length` should be set to 0 to disable IPv6 options (rarely used and non-zero attack surface)
- `ipv6.conf.$i.accept_redirects` should be 0, and this is sufficient.
- `ipv6.conf.$i.accept_ra` should be 0, and this is sufficient.
- `ipv6.conf.$i.accept_dad` should be 0, and this is *not* sufficient: `ipv6.conf.all.accept_dad` must also be 0.
- `ipv4.conf.$i.accept_source_route` should be 0, and this is sufficient.
- `ipv4.conf.$i.accept_redirects` should be 0 and this is *not* sufficient: `ipv4.conf.all.accept_redirects` must also be 0.
- It *might* be possible to use a sysctl file for this, but that is _not_ tested.  It would cover all `eth*` and `vif*` interfaces.

marmarek:
Since there are several unknowns about those settings, lets make this all opt-in (via qvm-service). If no issues will be reported, we can consider enabling it by default (maybe with opt-out?) in R4.2.

DemiMarie:
> Since there are several unknowns about those settings, lets make this all opt-in (via qvm-service). If no issues will be reported, we can consider enabling it by default (maybe with opt-out?) in R4.2.

I am fine with this, but this is technically tricky: it means that these settings need to be applied by the script, rather than by sysctl files.

DemiMarie:
> > I am fine with this, but this is technically tricky: it means that these settings need to be applied by the script, rather than by sysctl files.
> 
> Indeed, with standard sysctl config files you can't make it conditional. But you can still use `systemd-sysctl` and the same config format, just put the file in a different place, and load it from its own systemd service (that has appropriate condition on qvm-service).

That works.  I will give the the service that runs `setup-ip` `Requires=` and `After=` dependencies on this service.

DemiMarie:
PipelineRetry

network/setup-ip
```diff
@@ -238,6 +222,15 @@ if [ "$ACTION" == "add" ]; then
 
         primary_dns=$(/usr/bin/qubesdb-read /qubes-primary-dns 2>/dev/null) || primary_dns=
         secondary_dns=$(/usr/bin/qubesdb-read /qubes-secondary-dns 2>/dev/null) || secondary_dns=
+        /usr/lib/systemd/systemd-sysctl \
+            "--prefix=/net/ipv4/conf/all" \
+            "--prefix=/net/ipv4/neigh/all" \
+            "--prefix=/net/ipv6/conf/all" \
+            "--prefix=/net/ipv6/neigh/all"
```
marmarek, line 229: Missing `\`, making the whole script to fail.

network/vif-route-qubes
```diff
@@ -147,21 +147,15 @@ case $- in
...
-    echo 0 > "/proc/sys/net/ipv6/conf/$i/accept_dad"
-    echo 0 > "/proc/sys/net/ipv6/conf/$i/autoconf"
-    echo 1 > "/proc/sys/net/ipv6/conf/$i/drop_unicast_in_l2_multicast"
-    echo 1 > "/proc/sys/net/ipv6/conf/$i/drop_unsolicited_na"
-done
+/usr/lib/systemd/systemd-sysctl \
+    "--prefix=/net/ipv4/conf/all" \
+    "--prefix=/net/ipv4/neigh/all" \
+    "--prefix=/net/ipv6/conf/all" \
+    "--prefix=/net/ipv6/neigh/all"
```
marmarek, line 154: Here too

marmarek (changes requested):

DemiMarie:
PipelineRetryFailed

marmarek:
> Indeed, with standard sysctl config files you can't make it conditional. But you can still use `systemd-sysctl` and the same config format, just put the file in a different place, and load it from its own systemd service (that has appropriate condition on qvm-service).

This PR waits for the change requested above.

marmarek:
RPM build fails (outdated %files section).

marmarek (approved):

PR #360, opened 2022-02-07
Update keys with signatures that don't use SHA1 (release4.1)

HW42:
QubesOS/qubes-issues#6470

---

I'm not sure if the release4.1 and master branches have intentionally diverged already or if the release branch just hasn't been updated. If the later than this is a duplicate of PR #359.


marmarek:
Already backported.

PR #361, opened 2022-02-07
Update keys with signatures that don't use SHA1 (release4.0)

HW42:
QubesOS/qubes-issues#6470


marmarek:
cherry-picked from master already

PR #362, opened 2022-02-14
use common entry point for qvm-sync-clock

adrelanos:
fixes https://github.com/QubesOS/qubes-issues/issues/7265

### DIFF
--- a/qubes-rpc/suspend-post.d/qvm-sync-clock.sh
+++ b/qubes-rpc/suspend-post.d/qvm-sync-clock.sh
@@ -1,5 +1,6 @@
 #!/bin/sh
 
 if [ ! -f /var/run/qubes-service/clocksync ]; then
-    /usr/bin/qvm-sync-clock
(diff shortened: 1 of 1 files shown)


DemiMarie:
Signature check is failing because it expects the commit to be signed by your key, not GitHub’s.

adrelanos:
> Signature check is failing because it expects the commit to be signed by your key, not GitHub’s.

I noticed that but was hoping that is OK.

Are

* small / trivial changes
* below copyright threshold ([Threshold of originality](https://en.wikipedia.org/wiki/Threshold_of_originality))

(such as this) still acceptable using github web interface (github's key)?

marmarek:
> Are
> 
>     * small / trivial changes
> 
>     * below copyright threshold ([Threshold of originality](https://en.wikipedia.org/wiki/Threshold_of_originality))
> 
> 
> (such as this) still acceptable using github web interface (github's key)?

Yes, I'm fine with that.

marmarek (approved):

PR #363, opened 2022-02-15
qfile-agent: display GUI message only on PROGRESS_TYPE=gui

fepitre:
`kdialog` or `zenity` block the underlying process until we quit the popup. Either on server or automated workflows, we don't need GUI notifications that a requested file is missing.

### DIFF
--- a/qubes-rpc/qvm-copy
+++ b/qubes-rpc/qvm-copy
@@ -22,17 +22,18 @@ set -e
 #
 
 unset PROGRESS_TYPE OPERATION_TYPE TARGET_TYPE MIN_ARGS FILECOPY_TOTAL_SIZE
+
 case ${0##*/} in
-(qvm-move) OPERATION_TYPE=move TARGET_TYPE=default MIN_ARGS=1;;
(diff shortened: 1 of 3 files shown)


qubes-rpc/qvm-copy
```diff
@@ -22,41 +22,43 @@ set -e
...
+    else
+        echo "Copy FILE to ~/QubesIncoming/[THIS QUBE'S NAME]/ in the destination qube."
+    fi
 
-	echo
-	echo "You will be prompted to select the destination qube. If FILE is a directory, it will be copied recursively."
-	exit 1
+    echo
+    echo "You will be prompted to select the destination qube. If FILE is a directory, it will be copied recursively."
+    echo "In order to avoid typing target qube name twice, use qvm-copy/qvm-move instead of qvm-copy-to-vm/qvm-move-to-vm."
```
marmarek, line 60: This should be printed only for `$TARGET_TYPE = vm` case.

marmarek (changes requested):

qubes-rpc/qvm-copy
fepitre: I've added a check on that.

marmarek (approved):

PR #364, opened 2022-02-16
Disable CUPS by default

DemiMarie:
It is a security risk, slows qube boot, wastes memory and CPU, and is
not necessary.  Fixes QubesOS/qubes-issues#5179.

Reported-by: Patrick Schleizer <adrelanos@whonix.org>

@adrelanos I hope I got the Reported-by correct.

### DIFF
--- a/vm-systemd/qubes-sysinit.sh
+++ b/vm-systemd/qubes-sysinit.sh
@@ -7,7 +7,7 @@
 # List of services enabled by default (in case of absence of qubesdb entry)
 DEFAULT_ENABLED_NETVM="network-manager qubes-network qubes-update-check qubes-updates-proxy meminfo-writer qubes-firewall"
 DEFAULT_ENABLED_PROXYVM="qubes-network qubes-firewall qubes-update-check meminfo-writer"
-DEFAULT_ENABLED_APPVM="cups qubes-update-check meminfo-writer"
+DEFAULT_ENABLED_APPVM="qubes-update-check meminfo-writer"
 DEFAULT_ENABLED_TEMPLATEVM="$DEFAULT_ENABLED_APPVM updates-proxy-setup"
 DEFAULT_ENABLED="meminfo-writer"
 


marmarek (approved):

ninavizz:
Wait—so how will users print?? Sorry, this thread just showed-up on my radar... and, guys, users being able to print by default—or being able to _easily_ enable printing— is a pretty big deal. A LOT of users print! Not developers, no, but academics, researchers, enterprise/gov users... me! I only caught this PR because I recognized CUPS from SecureDrop stuff. Are there other print drivers/utilities that can do printing by default, that ship with Fedora/Debian?

While I agree with Patrick's flag and the necessity of Qubes OS honoring that reality, afaik CUPS is the only default print thingy in Qubes OS. This default ship-state of CUPS needs to be clearly documented, and at some point adding a "How to Print from Qubes OS" should be added to our docs that identifies CUPS as the Linux dealie that does printing; how it's enabled/disabled in Qubes, that Qubes templates have it disabled by default, etc.  /cc @andrewdavidwong

DemiMarie:
@ninavizz You are correct that this makes printing harder, sorry.  One _can_ do it via graphical interfaces, but I don’t think having to know that `cups` is the name of the service one must enable really counts, as there is zero discoverability.

ninavizz:
@DemiMarie I'm ok with making things "harder" against less-safe non-Qubes defaults, but only when there are alternate paths made very clear for a user to achieve that same goal. 

For the Mac and Windows user experience, things "just work." Linux, is famously not like that. Printing and network management, are two of the primary things users expect to "just work." Same, with files management. We don't have different driver apps to choose from to do those things with—and most of us are ok with that. That one might in Linux, has taken me a long time to adjust to (and I still, honestly, don't know what the point really is—especially since the CUPS GUI is one of the worst UIs I've ever seen).

Networking stuff is not entirely straightforward in Qubes, but at least is documented well. Printing—Macs and Windows just have printer apps. Non-geek Linux users, may even relate to CUPS like that, I don't know. 

TL;DR, yes, knowing that the name of the "how I can print on this machine" service, is very important for non-Linux native users to succeed, in Qubes. And, because of Qubes' superior security properties, it attracts non-Linux natives that are signing up to learn about Qubes, not Linux. 

We need to be more clear and transparent about how one directly leads into the other, while also providing clear/direct ways for users to do the most basic of things. While that may somewhat violate the "No redundant documentation" rule, when using Qubes requires hours of reading documentation to adapt to as a Mac or Windows user, that is an adoption problem I'm not comfortable burdening vulnerable users with.

andrewdavidwong:
> /cc @andrewdavidwong

Responded on https://github.com/QubesOS/qubes-issues/issues/7281.

marmarek:
There is no one place to enable CUPS everywhere, but you can enable CUPS
in specific qubes, either via command line (qvm-service --enable
QUBE-NAME cups), or via GUI (go to "services" tab in qube settings and
add "cups" service).

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

justusranvier:
Yeah I had already figured out there is a manual workaround, what I was hoping for was some file I could edit once to get the old behavior back instead.

The "printing no longer works unless you apply manual workarounds that are documented absolutely nowhere except github issue comments" feature really contributes to 4.2 being an incredibly cursed release.

q4747:
> The "printing no longer works unless you apply manual workarounds that are documented absolutely nowhere except github issue comments" feature really contributes to 4.2 being an incredibly cursed release.

See related discussion/guide here: https://forum.qubes-os.org/t/how-to-get-a-usb-mfp-printer-scanner-working-in-r4-2/24520

PR #367, opened 2022-03-12
Turn on scatter-gather and checksum offload

DemiMarie:
There is no reason to *not* have scatter-gather and checksum offload
turned on.

### DIFF
--- a/archlinux/PKGBUILD
+++ b/archlinux/PKGBUILD
@@ -35,7 +35,6 @@ build() {
     done
 
     # Fix for network tools paths
-    sed 's:/sbin/ethtool:ethtool:g' -i network/*
     sed 's:/sbin/ip:ip:g' -i network/*
     sed 's:/bin/grep:grep:g' -i network/*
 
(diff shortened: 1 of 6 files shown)


marmarek:
I'd rather remove those ethtool calls completely and rely on default (chosen by the kernel, or other network configuration tools in given distribution) - which are both "on" here.

Also, when rolling back some non-standard settings, it's helpful to do research why they were put there in the first place, to verify if really "There is no reason to not have [...]".

- checksum offload: https://github.com/QubesOS/qubes-issues/issues/700 (likely not relevant anymore, but it points at a configuration to test)
- scatter-gather: 5cbb38a2a92e93a1019f14d1f86524c8a61d608c - this is only about re-doing the change on dynamic netvm switch; unfortunately, the other call was added with initial commit for the startup scripts already; so, lets hope our tests cover relevant scenarios...

Finally, please add reference to https://github.com/QubesOS/qubes-issues/issues/3510 in the commit message

fepitre:
This is even more important because one of my physical network does not like to turn on SG. So don't enforce it at all is required.

DemiMarie:
> This is even more important because one of my physical network does not like to turn on SG. So don't enforce it at all is required.

Is it a problem if SG is enabled on the Xen interface?

marmarek (approved):

PR #369, opened 2022-03-24
Rip out the network-based updates proxy

DemiMarie:
It’s not used anymore and is useless attack surface.  Get rid of it.

Fixes QubesOS/qubes-issues#3285.

Not sure if this needs a test, since it just deletes code.


marmarek:
PipelineRefresh

marmarek (approved):

PR #371, opened 2022-04-10
Cohere with Debian guidance on 3rd party repos

unman:
Closes QubesOS/qubes-issues#7271


DemiMarie:
> Generally I like this change, but I worry about upgrade: if user modified the sources.list file (for example to enable testing updates), the new version will be saved in `.dpkg-new`, right? In that case, old file will be used, but the key in the old location will be gone. This will break installing any further updates.
> 
> Any ideas? Keep old keyring location (too) for the time being?

What about doing the upgrade in two steps?

PR #377, opened 2022-04-21
Rip out network-based update proxy remanents     

DemiMarie:
732e9d02eeae18c3cfe3c10beceed1564f28c399 removed most of the network-based updates proxy, but it didn't change the SysV init scripts.

### DIFF
--- a/vm-init.d/qubes-updates-proxy
+++ b/vm-init.d/qubes-updates-proxy
@@ -48,9 +48,6 @@ start() {
 
     [ -x "$exec" ] || exit 5
     [ -f $config ] || exit 6
-    # setup network redirection
-    /sbin/iptables -I INPUT -i vif+ -p tcp --dport 8082 -j ACCEPT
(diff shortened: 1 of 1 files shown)


vm-init.d/qubes-updates-proxy
```diff
@@ -42,17 +42,16 @@ start() {
...
     if qsvc qubes-updates-proxy ; then
-        # Yum proxy disabled
+        # Updates proxy disabled
         exit 0
     fi
 
     [ -x "$exec" ] || exit 5
     [ -f $config ] || exit 6
     # setup network redirection
     /sbin/iptables -I INPUT -i vif+ -p tcp --dport 8082 -j ACCEPT
```
marmarek, line 52: I think this line (and the matching one on stop) is not necessary either. The proxy is accessed over 'lo', not 'vif' interface.

marmarek (changes requested):

DemiMarie:
PipelineRetry

marmarek (approved):

PR #384, opened 2022-06-04
Use link scope for routes

DemiMarie:
Host scope means "only valid within this host", which is certainly not
correct for Xen virtual devices.  Link scope means "only valid in the
context of this link", which is correct.

Fixes: QubesOS/qubes-issues#7123.
Suggested-by: Ente <ducksource@duckpond.ch>

### DIFF
--- a/network/setup-ip
+++ b/network/setup-ip
@@ -5,8 +5,8 @@
 . /usr/lib/qubes/init/functions
 set -uo pipefail
 
-add_host_route () {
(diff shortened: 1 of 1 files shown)


Enteee:
@DemiMarie @marmarek : do you need me to do something about the failed tests?

DemiMarie:
PipelineRetry

DemiMarie:
PipelineRetry

DemiMarie:
PipelineRetry

marmarek (approved):

PR #386, opened 2022-06-13
Add support for /rw/config/rc.local-early to run commands as early as possible

chriswoope:
This allows to run arbitrary commands early in initialization, which is useful for things like replacing Qubes GUI packages in a specific VM.

### DIFF
--- a/vm-systemd/qubes-early-vm-config.sh
+++ b/vm-systemd/qubes-early-vm-config.sh
@@ -4,6 +4,10 @@
 # It happens after local-fs.target is reached
 # but before sysinit.target is reached.
 
+if [ -x /rw/config/rc.local-early ] ; then
+    /rw/config/rc.local-early
+fi
+
 # Source Qubes library.
 # shellcheck source=init/functions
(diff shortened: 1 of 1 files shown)


marmarek:
@tasket will that interfere with https://github.com/tasket/Qubes-VM-hardening ? On the first sight, it shouldn't but I'd like to double check.

3hhh:
> This allows to run arbitrary commands early in initialization, which is useful for things like replacing Qubes GUI packages in a specific VM.

Just use systemd for that?

marmarek:
> Just use systemd for that?

You don't have (early enough) place to register systemd units stored in `/rw`. This addition could be such place.

ben-grande:
If this is ever accepted, please remember to add drop-in configuration directory `/rw/config/rc.local-early.d` to follow packaging standards of allowing multiple configurations.

FijiHasGithub:
Any news on this? Would be really helpful with a problem I am currently having

marmarek (approved):

PR #389, opened 2022-07-15
Allow systemd to manage PulseAudio

DemiMarie:
This allows systemd to manage PulseAudio.  Instead of marking the
PulseAudio service as broken, override the `ExecStart=` command.

### DIFF
--- a/vm-systemd/user/pulseaudio.service.d/30_qubes.conf
+++ b/vm-systemd/user/pulseaudio.service.d/30_qubes.conf
@@ -1,4 +1,7 @@
 [Unit]
-#ConditionNull=false
-ConditionPathExists=/var/run/qubes-service/pulseaudio-vanilla-broken
+ConditionPathExists=/etc/pulse/qubes-default.pa
 
+[Service]
+ExecStartPre=-/usr/bin/qubesdb-read -w /qubes-audio-domain-xid
+ExecStart=
+ExecStart=/usr/bin/pulseaudio --start -n --file=/etc/pulse/qubes-default.pa --exit-idle-time=-1 --daemonize=no --log-target=journal
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ SYSTEM_DROPINS += systemd-logind.service
 SYSTEM_DROPINS_NETWORKING := NetworkManager.service NetworkManager-wait-online.service
 SYSTEM_DROPINS_NETWORKING += tinyproxy.service
 
-USER_DROPINS := pulseaudio.service pulseaudio.socket
+USER_DROPINS := pulseaudio.service
 
 # Ubuntu Dropins
 ifeq ($(release),Ubuntu)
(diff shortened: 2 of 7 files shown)


marmarek:
Does it require disabling `start-pulseaudio-with-vchan`? Or maybe at least making it no-op if systemd will start pulseaudio?

marmarek:
Are you sure that all (relevant) distributions that use systemd, starts pulseaudio via systemd user unit?

DemiMarie:
> Are you sure that all (relevant) distributions that use systemd, starts pulseaudio via systemd user unit?

Not at all.  I will add a check for `/lib/systemd/user/pulseaudio.service` existing.

DemiMarie:
PipelineRetry

vm-systemd/user/pulseaudio.service.d/30_qubes.conf
```diff
@@ -1,4 +1,6 @@
 [Unit]
-#ConditionNull=false
-ConditionPathExists=/var/run/qubes-service/pulseaudio-vanilla-broken
+ConditionPathExists=/etc/pulse/qubes-default.pa
 
+[Service]
```
marmarek, line 4: The service needs also take care about waiting for audiovm to start: https://github.com/QubesOS/qubes-gui-agent-linux/commit/cabf69ffb8abacac0c12ef929637f5e223f4a1dd#diff-3c748b2e4495de699851b8c7b2144314e463fd40f14e04453c60c8fb87fa86c6R30 (and previous commit about that part). Alternatively, this can be moved into pulseaudio plugin itself. I guess PipeWire agent needs similar part too.

marmarek (changes requested):

vm-systemd/user/pulseaudio.service.d/30_qubes.conf
DemiMarie: For PipeWire my plan is to do what I did with Wayland, which is to treat every connection as a reconnection.  This avoids having to wait for the audio daemon to connec.

marmarek: Regardless of the final solution (the above idea sounds fine), please do the minimum required to not regress functionality here. If moving the waiting to the pulseaudio plugin itself is simple enough, go for it, otherwise add waiting as `ExecStartPre` or similar (in that case verify interaction with various timeouts - better to assume it may take at least few minutes to start).

DemiMarie: I don’t see any timeouts.

DemiMarie:
PipelineRetryFailed

marmarek:
In fact it's weird to have this drop-in file in this package, while the rest of pulseaudio is elsewhere. I'll move it and see what happens.

marmarek:
Done in #401 instead.

PR #398, opened 2022-11-30
qubes-rpc/nautilus: Execute external commands asynchronously

noskb:
Using [GLib.spawn_async](https://lazka.github.io/pgi-docs/#GLib-2.0/functions.html#GLib.spawn_async) instead of subprocess.call or subprocess.Popen 

This prevents Nautilus from getting stuck while executing external commands. And no zombies.
As far as I've tried, Nautilus 43 on fedora-37 gets stuck without closing the context menu when calling external commands synchronously, so it should also help prevent that.

I tested with debian-10 and fedora-37 templates.

### DIFF
--- a/qubes-rpc/nautilus/qvm_copy_nautilus.py
+++ b/qubes-rpc/nautilus/qvm_copy_nautilus.py
@@ -1,6 +1,4 @@
-import subprocess
-
-from gi.repository import Nautilus, GObject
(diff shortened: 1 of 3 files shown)


marmarek (approved):

andrewdavidwong:
Fixes QubesOS/qubes-issues#8006

PR #404, opened 2023-01-19
qvm-connect-tcp: print usage on invalid parameters

meithecatte:
This also handles things like `qvm-connect-tcp --help`, which are otherwise met with a very unhelpful error message.


marmarek:
PipelineRetry

meithecatte:
I might be wrong, but the CI failures seem to be unrelated to the actual change.

marmarek (approved):

PR #406, opened 2023-01-30
network: fix uplink setup relation to network.target

rustybird:


rustybird:
One remaining (but unrelated to this PR) failed job in `install:vm-bookworm`

    The following packages have unmet dependencies:
     python3-qubesdb : Depends: python3 (< 3.11) but 3.11.1-2 is to be installed

marmarek (approved):

PR #407, opened 2023-02-05
Switch to nftables

DemiMarie:
Draft as packaging has not been updated.

Fixes: QubesOS/qubes-issues#5031
Fixes: QubesOS/qubes-issues#6062


network/setup-ip
```diff
@@ -188,36 +196,28 @@ fi
...
+    if [[ -n "$MAC" ]]; then
+        custom=$(exec /usr/bin/qubesdb-exists "${prefix}custom")
+        if ip4=$(exec /usr/bin/qubesdb-read "${prefix}ip" 2>/dev/null); then
+            :
+        elif [[ "$?" != '2' ]]; then
+            echo "Could not check if /net-config/$MAC/ip exists!" >&2
+            exit 1
+        else
+            prefix=/qubes-
+            ip4=$(exec /usr/bin/qubesdb-read "${prefix}ip")
```
marmarek, line 209: shellcheck complains this variable is unused, which means you haven't updated its name down below.

network/setup-ip
```diff
@@ -188,36 +196,28 @@ fi
...
+        netmask=$(exec /usr/bin/qubesdb-read -e 255.255.255.255 "${prefix}netmask")
+        gateway=$(exec /usr/bin/qubesdb-read "${prefix}gateway")
+        if ip6=$(exec /usr/bin/qubesdb-read "${prefix}ip6" 2>/dev/null); then
+            netmask6=$(exec /usr/bin/qubesdb-read -e 128 "${prefix}netmask6")
+            gateway6=$(exec /usr/bin/qubesdb-read "${prefix}gateway6")
+        elif [[ "$?" != '2' ]]; then
+            echo 'Could not check if IPv6 is enabled' >&2
+            exit 1
+        else
+            ip6= netmask6=128 gateway6=
```
marmarek, line 220: shellcheck suggests `ip6=""` and I agree.

marmarek (changes requested):

network/qubes-iptables
```diff
@@ -21,36 +21,29 @@
...
+if [ ! -x "/sbin/$IPTABLES" ]; then
     echo $"${IPTABLES}: /sbin/$IPTABLES does not exist."
     exit 5
 fi
 
-start() {
+start () {
     ipt=$1
-    IPTABLES_DATA=$IPTABLES_DATA_DIR/${ipt}.rules
+    IPTABLES_DATA=$IPTABLES_DATA_DIR/qubes-${ipt}.rules
```
marmarek, line 31:
```suggestion
    IPTABLES_DATA=$IPTABLES_DATA_DIR/qubes-${ipt}.nft
```
?

The service fails to start with code 6 (unfortunately no more info is logged), I guess because of this (and the other .rules below).

Then, vif-route-qubes fails too, because it can't find `qbs-raw` table.

DemiMarie: :woman_facepalming: oops, I don’t use sysvinit so this got no testing.

marmarek: > woman_facepalming oops, I don’t use sysvinit so this got no testing.

You don't need to. This script is called by `qubes-iptables.service`.

network/setup-ip
```diff
@@ -188,36 +209,36 @@ fi
...
+        elif [[ "$?" != '2' ]]; then
+            echo "Could not check if /net-config/$MAC/ip exists!" >&2
+            exit 1
+        else
+            prefix=/qubes-
+            ip4=$(exec /usr/bin/qubesdb-read "${prefix}ip")
         fi
-        if [ -z "$netmask6" ]; then
-            netmask6="128"
+        netmask=$(exec /usr/bin/qubesdb-read -e 255.255.255.255 "${prefix}netmask")
```
marmarek, line 232: `qubesdb-read` doesn't have `-e` option (few instances below too), is it some your local change? Looks like "default value", which sounds like a useful feature, but I'm not sure if `-e` is the most intuitive option for that, I'd rather choose `-d`.

marmarek (changes requested):

marmarek:
`setup-ip` fails with:
```
iptables: No chain/target/match by that name.
iptables-restore: line 6 failed
```
My guess is https://github.com/QubesOS/qubes-core-agent-linux/blob/71d2fb884451eea25411c7e589c02f9eff02adc3/network/qubes-setup-dnat-to-ns called from setup-ip

marmarek:
PipelineRetry

network/qubes-setup-dnat-to-ns
```diff
@@ -1,34 +1,99 @@
...
+        if ns_maybe is None:
+            continue
+        try:
+            qubesdb_dns.append(IPv4Address(ns_maybe.decode("ascii", "strict")))
+        except (UnicodeDecodeError, ValueError):
+            pass
+    res = [
+        'add table ip qubes',
+        'add chain ip qubes dnat-dns',
+        'delete chain ip qubes dnat-dns',
```
marmarek, line 85: This could use a comment why.

network/vif-route-qubes
```diff
@@ -173,29 +194,44 @@ fi
...
+                dev "$vif" lladdr "$peer_mac" nud permanent
+        fi
+        nft "$nftables_cmd element $ipt qbs-raw allowed { $ifindex . $addr }
+$nftables_cmd element $ipt qbs-raw downstream { $addr }"
     done
     # if no IPv6 is assigned, block all IPv6 traffic on that interface
     if ! [[ "$ip" = *:* ]]; then
-        echo -e "*raw\\n$iptables_cmd -i ${vif} -j DROP\\nCOMMIT" |
-            ${cmdprefix} ip6tables-restore --noflush $ipt_arg
+        echo 1 > "/proc/sys/net/conf/$vif/disable_ipv6"
```
marmarek, line 234: AFAIR there was a reason why it uses firewall, not disable_ipv6 flag. I don't remember details but it could be some applications failing when they prefer IPv6, or when using `::IPv4` style addresses.

marmarek (changes requested):

marmarek:
In the ongoing test run I see failures like this in /var/log/xen/xen-hotplug.log:
```
Error: No such file or directory
add element ip qbs-raw allowed { 26 . 10.137.0.17 }
               ^^^^^^^
```
Looks like you forgot to update table name.

marmarek:
> * VmNetworking_debian-11: [test_000_simple_networking](https://openqa.qubes-os.org/tests/71193#step/VmNetworking_debian-11/1) (failure)
>       `dbus.exceptions.DBusException: org.freedesktop.systemd1.NoSuchUnit:...`

And also this.

marmarek:
All networking fails with `dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NameHasNoOwner: Could not get owner of name 'org.freedesktop.resolve1': no such name`
I don't see any obvious typo, but also the fallback to /etc/resolv.conf didn't kicked in.

marmarek:
The current error handling looks fragile there, I don't think you have a guarantee those errors covers all the cases that should result in a fallback. And even if it does now, it looks like it may change or other distros can have yet another behavior. Maybe in case of a dbus error, check if systemd-resolved service is running and decide on fallback/fail based on that?

network/vif-route-qubes
```diff
@@ -20,6 +20,7 @@
 #         this script).
 #============================================================================
 
+set -e
```
marmarek, line 23: I'm not sure if this only uncovered some pre-existing issue (more likely), or if there is a bug in this PR, but vif-qubes-nat.sh now fails. The end effect is that with this PR at least masquerading IP doesn't work anymore. The tests hanged, so I don't have full log, but I have a screenshot with some of the details (see the bottom part): https://openqa.qubes-os.org/tests/71340#step/system_tests/19

Another thing is most networking test fail with IPv6 enabled (https://openqa.qubes-os.org/tests/71351), but that could be  an issue with the tests themselves (given I found https://github.com/QubesOS/qubes-core-admin/pull/528 already). I'll analyze test logs a bit more.

marmarek:
> [unstable] VmNetworking_debian-11: [test_020_simple_proxyvm_nm](https://openqa.qubes-os.org/tests/71473#step/VmNetworking_debian-11/3) (failure)
AssertionError: 1 != 0 : Ping by IP failed

This is real issue, NetworkManger enabled in proxy qube tried to configure the network using DHCP. Something went wrong with providing it static config. Specifically this:
> [2023-04-20 08:34:07] [   20.724089] setup-ip[574]: /usr/lib/qubes/setup-ip: line 91: 11: unbound variable

This, and the earlier issue about qubes-nat.sh are the only remaining failures obviously related to this PR.

DemiMarie:
> > [unstable] VmNetworking_debian-11: [test_020_simple_proxyvm_nm](https://openqa.qubes-os.org/tests/71473#step/VmNetworking_debian-11/3) (failure)
> > AssertionError: 1 != 0 : Ping by IP failed
> 
> This is real issue, NetworkManger enabled in proxy qube tried to configure the network using DHCP. Something went wrong with providing it static config. Specifically this:
> 
> > [2023-04-20 08:34:07] [   20.724089] setup-ip[574]: /usr/lib/qubes/setup-ip: line 91: 11: unbound variable
> 
> This, and the earlier issue about qubes-nat.sh are the only remaining failures obviously related to this PR.

Both fixed.

DemiMarie:
PipelineRetryFailed

marmarek:
@DemiMarie I added appropriate fixes tags to the PR description, but please also add them to relevant commits.

marmarek:
There is one more thing, Whonix Gateway is broken:
```
[2023-04-21 05:20:41] [  230.180382] root[1377]: /etc/xen/scripts/vif-route-qubes: /etc/xen/scripts/vif-route-qubes failed; error detected.
```
I don't have xen-hotplug.log saved, but I'm pretty sure it's because Whonix replaces `qubes-iptables` service with its own thing. I see three solutions here:
1. Modify vif-route-qubes to keep using iptables on Whonix  (IMO least preferred).
2. Modify Whonix's firewall to use nft _and_ include qubes-specific chains for antispoofing rules (while the first part might be the desirable long term, not sure about the second part).
3. Split creating antispoofing nft chains to a separate service that that is not going to be disabled on Whonix and do not interfere with its firewall (really does nothing more than blocking IP spoofing attempts). IMO this is the most preferred option.

@adrelanos do you have any preference about this? For context, this is https://github.com/QubesOS/qubes-issues/issues/5031, which is going to be included in R4.2. Schedule-wise, first rc will be released "soon"(TM), but the final release most likely happen after Debian bookworm release. I'm not sure how that aligns with Whonix 17 release schedule, if that would be relevant here.

DemiMarie:
PipelineRetryFailed

marmarek:
Please drop the selinux change, there is an alternative: https://github.com/QubesOS/qubes-core-agent-linux/pull/417

marmarek:
Besides the Whonix compat issue, the "fake IP" still fails:
```
Apr 26 22:22:22 test-inst-netvm1 root[1080]: /etc/xen/scripts/vif-route-qubes: online type_if=vif XENBUS_PATH=backend/vif/41/0
Apr 26 22:22:23 test-inst-netvm1 root[1116]: /etc/xen/scripts/vif-route-qubes: /etc/xen/scripts/vif-route-qubes failed; error detected.
...
cat /var/log/xen/xen-hotplug.log: b'Cannot remove namespace file "/run/netns/vif41.0-nat": No such file or directory\n/etc/xen/scripts/vif-qubes-nat.sh: line 51: /proc/sys/net/ipv6/conf//disable_ipv6: No such file or directory\n'
```

DemiMarie:
> Besides the Whonix compat issue,

Should this be part of this or a separate PR?

> the "fake IP" still fails:
> 
> ```
> Apr 26 22:22:22 test-inst-netvm1 root[1080]: /etc/xen/scripts/vif-route-qubes: online type_if=vif XENBUS_PATH=backend/vif/41/0
> Apr 26 22:22:23 test-inst-netvm1 root[1116]: /etc/xen/scripts/vif-route-qubes: /etc/xen/scripts/vif-route-qubes failed; error detected.
> ...
> cat /var/log/xen/xen-hotplug.log: b'Cannot remove namespace file "/run/netns/vif41.0-nat": No such file or directory\n/etc/xen/scripts/vif-qubes-nat.sh: line 51: /proc/sys/net/ipv6/conf//disable_ipv6: No such file or directory\n'
> ```

Fixed.

marmarek:
> > Besides the Whonix compat issue,
> 
> Should this be part of this or a separate PR?

Yes, this one, otherwise this PR regresses Whonix support.

DemiMarie:
PipelineRetryFailed

marmarek:
No network traffic is passed through, `/var/log/xen/xen-hotplug.log`:
```
Error: No such file or directory
add element ip qubes allowed { "vif3.0" . 10.138.16.176 }
```

First issue is that `qubes-antispoof.service` is not started, because it's neither enabled via preset file, nor pulled in via dependency (`qubes-firewall.service` references wrong name - `qubes-iptables-antispoof.service`). But even after starting it manually, it still doesn't work, because `qubes-antispoof.service` actually starts the same `qubes-iptables` script (and so, the same nft rules), instead of ` qubes-antispoof.nft`. Loading `qubes-antispoof.nft` manually fixes the issue.

vm-systemd/qubes-iptables.service
```diff
@@ -1,5 +1,7 @@
 [Unit]
 Description=Qubes base firewall settings
+Requires=qubes-iptables.service
+After=qubes-iptables.service
```
marmarek, line 4: Uhm, now it references itself...

But also, I think the `qubes-antispoof.service` should be included in the preset file too, so it would be started even if somebody disables `qubes-iptables.service` (via systemctl, not qvm-service), as `qubes-iptables.service` shouldn't really be necessary for proper `vif-route-qubes` function.

marmarek (changes requested):

DemiMarie:
> > system_tests_network
> > VmNetworking_debian-11: [test_111_dynamic_detach_attach](https://openqa.qubes-os.org/tests/72785#step/VmNetworking_debian-11/9) (failure)
> > self.assertEqual(self.run_cmd(self.testvm1, ... AssertionError: 2 != 0
> 
> This looks like a legitimate failure, happens for both debian and fedora. The test starts the VM with netvm attached, detaches it and the attaches it again. I'm not sure what specifically failed there, but logs say `eth0` is not configured at the end (it isn't "up"). I haven't found specific error in the logs. Note the test when VM starts with netvm detached (set to "none") and is attached only after starting the VM passes.

This is actually a race condition in the test.  I can reproduce this with normal `qvm-run` commands if I add `sleep 5` (smaller amounts might also work) to `/usr/lib/qubes/setup-ip` just before bringing up the interface.  With the script in this PR things work for me.  https://github.com/QubesOS/qubes-core-admin/pull/532 should fix it.

DemiMarie:
QubesOS/qubes-core-admin#533 fixes the testsuite race

adrelanos:
> 1. Modify vif-route-qubes to keep using iptables on Whonix  (IMO least preferred).

Not ideal but OK with me.

> 2. Modify Whonix's firewall to use nft _and_ include qubes-specific chains for antispoofing rules (while the first part might be the desirable long term, not sure about the second part).

For porting Whonix to nft there's Whonix ticket [Consider nftables / Berkeley Packet Filter (BPF) as a replacement for iptables](https://phabricator.whonix.org/T509). But not easy at all. Critical task. Could use help with this one.

> 3. Split creating antispoofing nft chains to a separate service that that is not going to be disabled on Whonix and do not interfere with its firewall (really does nothing more than blocking IP spoofing attempts). IMO this is the most preferred option.

Also sounds good.

> @adrelanos do you have any preference about this?

> For context, this is [QubesOS/qubes-issues#5031](https://github.com/QubesOS/qubes-issues/issues/5031), which is going to be included in R4.2. Schedule-wise, first rc will be released "soon"(TM), but the final release will most likely happen after Debian bookworm release. I'm not sure how that aligns with Whonix 17 release schedule, if that would be relevant here.

Whonix 17 development (port to Debian bookworm) will start after Debian bookworm release.

marmarek (approved):
While there are still few network-related issues (QubesOS/qubes-issues#8178, QubesOS/qubes-issues#8169, https://github.com/QubesOS/qubes-core-admin/pull/533), there doesn't seem to be any remaining issue in the changes introduced here. Letslets handle remaining issue in a follow-up PR(s)

PR #414, opened 2023-04-20
Do not hard-code `user`

DemiMarie:
Qubes VMs can have any default user permitted by the operating system, provided that it can fit in qubesd.

### DIFF
--- a/debian/qubes-core-agent.postinst
+++ b/debian/qubes-core-agent.postinst
@@ -149,7 +149,7 @@ case "${1}" in
         fi
         systemctl reenable haveged || :
 
-        chgrp user /var/lib/qubes/dom0-updates
+        chgrp qubes /var/lib/qubes/dom0-updates
 
         debug "UPDATE..."
         # disable some Upstart services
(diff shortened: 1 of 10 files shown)


debian/qubes-core-agent.postinst
```diff
@@ -186,8 +186,8 @@ case "${1}" in
             rm /etc/systemd/system/multi-user.target.wants/qubes-mount-home.service
         fi
 
-        if ! dpkg-statoverride --list /var/lib/qubes/dom0-updates >/dev/null 2>&1; then
-            dpkg-statoverride --update --add user user 775 /var/lib/qubes/dom0-updates
+        if ! dpkg-statoverride --list /var/lib/qubes/dom0-updates >/dev/null 2>&1 && user=$(qubesdb-read /default-user); then
```
marmarek, line 189: This needs to have a default to `user`, qubesdb may not function during package install yet (when building fresh template).

marmarek: ping?

marmarek (approved):

PR #425, opened 2023-05-10
Remount /lib/modules in early boot

DemiMarie:
This unbreaks writing to /lib/modules with SELinux enforcing.

Reported-by: Marek Marczykowski-Górecki <marmarek@invisiblethingslab.com>
Fixes: QubesOS/qubes-issues#8144

### DIFF
--- a/debian/qubes-core-agent.install
+++ b/debian/qubes-core-agent.install
@@ -88,6 +88,7 @@ lib/systemd/system/qubes-mount-dirs.service
 lib/systemd/system/qubes-rootfs-resize.service
 lib/systemd/system/qubes-sysinit.service
 lib/systemd/system/qubes-update-check.service
+lib/systemd/system/qubes-remount-lib-modules.service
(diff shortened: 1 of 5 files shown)


marmarek:
This needs at least to check somewhere if it's actually using dom0-provided kernel. Checking for `/dev/xvdd` existence should work, but that's probably not the only option. Or maybe, it should check if a) SELinux is enabled, and b) `/lib/modules` (or a subdirectory of it?) has a wrong label? IOW, if remounting is necessary.

marmarek:
And also, shellcheck complains.

DemiMarie:
`ConditionSecurity=SELinux` and `ConditionKernelVersion=*.qubes.*` should do the job.

marmarek:
> Fixes: [QubesOS/qubes-issues#8144](https://github.com/QubesOS/qubes-issues/issues/8144)

I can confirm it fixes this issue indeed.
There are a couple of failed tests, but probably unrelated to this PR (needs to double check).

PR #429, opened 2023-05-24
Label Xen hotplug scripts correctly

DemiMarie:
They should be `initrc_exec_t`, but due to an incorrect regular expression this did not happen.  `[!/]+` matches one or more `!` or `/` characters, whereas `[^/]+` matches one or more characters that are not `/`.


marmarek:
that won't fix existing labels on package update, right? or maybe `%selinux_relabel_post` macro takes care of that?

marmarek:
> that won't fix existing labels on package update, right? or maybe `%selinux_relabel_post` macro takes care of that?

Looking at the macro implementation, I think it only will work if `%selinux_relabel_pre` is used too (which currently isn't).

PR #432, opened 2023-05-31
Use network-pre.target to order firewall before network uplink

DemiMarie:
network-pre.target is intended to be ordered after any service that sets up a firewall and before any service that enables networking. Therefore, have network-pre.target require qubes-antispoof.service (unconditionally) and qubes-iptables.service (only if qubes-iptables.service is enabled).  qubes-network-uplink@.service now has Requires= and After= dependencies on network-pre.target, so it is ordered after both Qubes and third-party firewall rules.

### DIFF
--- a/vm-systemd/qubes-antispoof.service
+++ b/vm-systemd/qubes-antispoof.service
@@ -4,7 +4,8 @@ Description=Qubes anti-spoofing firewall rules
 [Service]
 Type=oneshot
 RemainAfterExit=yes
+Before=network-pre.target
 ExecStart=/usr/sbin/nft -f /etc/qubes/qubes-antispoof.nft
 
 [Install]
-WantedBy=basic.target
+RequiredBy=network-pre.target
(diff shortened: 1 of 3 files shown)


marmarek:
there is a file conflict

vm-systemd/network-pre.target.d/30_qubes.conf
```diff
@@ -0,0 +1,3 @@
+[Unit]
+Requires=qubes-antispoof.service
+After=qubes-antispoof.service
```
marmarek, line 3: I think it's better to add them to the `qubes-antispoof.service` instead (as `RequiredBy=` and `Before=` respectively).

marmarek (changes requested):

marmarek:
Does this by a chance helps with https://github.com/QubesOS/qubes-issues/issues/8305 too?

marmarek (approved):

DemiMarie:
> Does this by a chance help with [QubesOS/qubes-issues#8305](https://github.com/QubesOS/qubes-issues/issues/8305) too?

I have not checked, but I would not at all be surprised either, since it removes the dependency on `sys-subsystem-net-device-%i.device`.

DemiMarie:
> Does this by a chance help with [QubesOS/qubes-issues#8305](https://github.com/QubesOS/qubes-issues/issues/8305) too?

It does indeed!  At first I thought it did not, but that turned out to be a case of PEBKAC.

PR #436, opened 2023-06-30
Do not fail if address already exists

DemiMarie:
This patch fixes fairly frequent network uplink failures on my machine.

Fixes: QubesOS/qubes-issues#8305

### DIFF
--- a/network/setup-ip
+++ b/network/setup-ip
@@ -27,13 +27,13 @@ configure_network () {
     local secondary_dns="${10}"
     local custom="${11}"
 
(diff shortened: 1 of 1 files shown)


marmarek:
As noted in the ticket, if the script it called on already configured interface, there are more things that will likely fail, for example adding routing entries. But I wonder why `setup-ip` would be called twice?

DemiMarie:
@marmarek because PEBKAC :rofl, closing.

DemiMarie:
Reopening because of QubesOS/qubes-issues#8362.

PR #437, opened 2023-07-01
Set correct SELinux contexts on /rw

DemiMarie:
This is needed for disposable sys-net to work properly.  Without it /rw is not labeled correctly, causing SELinux to (correctly) block NetworkManager's writes to /rw/config/NM-system-connections.

Fixes: QubesOS/qubes-issues#8242

### DIFF
--- a/init/functions
+++ b/init/functions
@@ -178,19 +178,29 @@ initialize_home() {
     home_root="$1"
     mode="$2"
 
-    if [ -z "$home_root" ] ; then
+    case $home_root in
+    (/*)
+        ;;
+    ('')
         echo "initialize_home() needs a target home root directory, such as /rw/home, as first parameter" >&2
         return 64
-    fi
+        ;;
+    (*)
+        echo 'initialize_home target root home directory must be an absolute path' >&2
+        return 64
+        ;;
+    esac
 
     if [ "$mode" != "unconditionally" ] && [ "$mode" != "ifneeded" ] ; then
         echo "initialize_home() second parameter must be 'unconditionally' or 'ifneeded'" >&2
         return 64
     fi
 
+    if test -d /sys/fs/selinux; then enable_selinux="Z"; else enable_selinux=''; fi
+
     if ! [ -d "$home_root" ] ; then
         echo "initialize_home: populating $home_root" >&2
-        mkdir -p "$home_root"
+        mkdir "-${enable_selinux}m0755" -- "$home_root" || return 73
     fi
 
     # Chown home if users' UIDs have changed - can be the case on template switch.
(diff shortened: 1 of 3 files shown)


marmarek:
Isn't it better to simply make `qubes-relabel-rw.service` working in DispVM too (drop its condition on persistence)?

DemiMarie:
> Isn't it better to simply make `qubes-relabel-rw.service` working in DispVM too (drop its condition on persistence)?

yes, but it is also better to create the files with the correct contexts in the first place.

marmarek:
In that case, `network/network-manager-prepare-conf-dir` may need adjustment too.

BTW, shellcheck complains.

marmarek:
ping?

DemiMarie:
fixed, assuming shellcheck is smart enough to note that `${SOMEVAR:+-Z}` without quoting is perfectly safe.

init/functions
```diff
@@ -202,8 +212,14 @@ initialize_home() {
...
             echo "initialize_home: populating $mode $home_root/$homedirwithouthome from /etc/skel" >&2
-            mkdir -p "$home_root/$homedirwithouthome"
-            if test -d /sys/fs/selinux; then enable_selinux="Z"; else enable_selinux=''; fi
+            if [ "$mode" = unconditionally ]; then
+                mkdir "-p${enable_selinux}" -- "$home_root/$homedirwithouthome" || return 73
+            else
+                case $homedirwithouthome in
+                (*/*) mkdir "-p${enable_selinux}" -- "$home_root/${homedirwithouthome%/*}";;
+                esac
+                mkdir "${enable_selinux+-Z}" -- "$home_root/$homedirwithouthome" || return 73
```
marmarek, line 221: Why splitting this into two mkdir calls (for which you needed to split this into two branches and extra pattern matching)? If you want this to fail if directory already exists, there is no point in that - this part is under `if ! test -d "$home_root/$homedirwithouthome"` (and this script doesn't really need to worry about racing against anything else accessing `/rw`).
But if there is some other reason, please explain it in the commit message.

init/setup-rw.sh
```diff
@@ -13,11 +13,18 @@ if mountpoint -q /rw ; then
         echo "Private device size management: resize2fs $dev failed:" >&2
         echo "$content" >&2
     fi
+    unset enable_selinux
```
marmarek, line 16: The unset isn't really necessary, if you set it unconditionally anyway.

init/setup-rw.sh
```diff
@@ -55,15 +62,17 @@ EOF
 # file is used only if the VM has any PCI device assigned. Modules will be
 # automatically re-loaded after resume.
 EOF
+        if [ -n "$enable_selinux" ]; then restorecon -RF /rw/config; fi
```
marmarek, line 65: `qubes-relabel-rw.service` (which is now also enabled in DispVMs) will do that again. if `restorecon` is called here (and in few other cases below), maybe create `/rw/.autorelabel` too (which BTW is misnamed...).

marmarek (changes requested):

marmarek (approved):
I'll regret accepting this blown up mkdir but okay...

PR #440, opened 2023-07-10
Drop requirement on iptables

DemiMarie:
iptables is not used any more.

Fixes: QubesOS/qubes-issues#5301
Fixes: QubesOS/qubes-issues#8346


DemiMarie:
PipelineRetryFailed

PR #441, opened 2023-07-15
Do not fail if address already exists

DemiMarie:
This patch fixes network uplink failures on Fedora 37 and Debian 11 VMs with only one vCPU.  There is an underlying problem that causes the script to be run more than once, but this patch is a workaround until that bug can be found and fixed.

Fixes: QubesOS/qubes-issues#8305
Fixes: QubesOS/qubes-issues#8362

### DIFF
--- a/network/setup-ip
+++ b/network/setup-ip
@@ -27,13 +27,13 @@ configure_network () {
     local secondary_dns="${10}"
     local custom="${11}"
 
-    /sbin/ip -- address add "$ip/$netmask" dev "$INTERFACE"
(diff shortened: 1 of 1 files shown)


marmarek:
If we really need to go this way, then the same comment as in previous attempt applies:
> if the script it called on already configured interface, there are more things that will likely fail, for example adding routing entries

DemiMarie:
> If we really need to go this way, then the same comment as in previous attempt applies:
> 
> > if the script it called on already configured interface, there are more things that will likely fail, for example adding routing entries

This change alone is enough to make things work for me, so this does not seem to be a problem in practice.

marmarek (approved):

PR #447, opened 2023-07-26
Make services conditional

DemiMarie:
This allows services that are currently started in every non-minimal Fedora qube to be controlled via `qvm-service`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,13 @@ endif
 SYSTEM_DROPINS_NETWORKING := NetworkManager.service NetworkManager-wait-online.service
 SYSTEM_DROPINS_NETWORKING += tinyproxy.service
 
-USER_DROPINS :=
+USER_DROPINS := \
+	tracker-xdg-portal-3.service \
+	tracker-writeback-3.service \
+	tracker-miner-rss-3.service \
+	tracker-miner-fs-control-3.service \
+	tracker-miner-fs-3.service \
+	tracker-extract-3.service
 
 # Ubuntu Dropins
 ifeq ($(release),Ubuntu)
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,11 @@ USER_DROPINS := \
 	tracker-miner-rss-3.service \
 	tracker-miner-fs-control-3.service \
 	tracker-miner-fs-3.service \
-	tracker-extract-3.service
+	tracker-extract-3.service \
+	evolution-addressbook-factory.service \
+	evolution-calendar-factory.service \
+	evolution-source-registry.service \
+	evolution-user-prompter.service
 
 # Ubuntu Dropins
 ifeq ($(release),Ubuntu)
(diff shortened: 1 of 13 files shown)


marmarek:
First of all, this doesn't build (spec file needs an update).
But then, it needs proper issue for tracking and including in release notes. And an update to qvm-service man page (in core-admin-client repo).
And also, a **really good** argument why it's worth an exception from https://www.qubes-os.org/faq/#what-is-qubes-attitude-toward-changing-guest-distros. Lacking that, the default would need to stay unmodified (enabled).

DemiMarie:
- Thunderbird downloads untrusted email attachments to `~/Downloads`, which then get parsed by Tracker.  If there is an RCE in Tracker, this could be exploited even if the user would have only viewed the file in a disposable VM.
- Evolution Data Server almost never has any data to serve.  The default email client in Qubes is Thunderbird, not Evolution, and this means that Evolution Data Server never gets any useful data.
- Both Tracker and Evolution Data Server are tightly integrated with GNOME, but Qubes OS does not use GNOME.  KDE uses its own searcher (Baloo) and I am not aware of XFCE using one at all.
- Tracker uses 61.8MB of memory in one of my disposable VMs, even though it is not doing anything useful!  61.8MB × 10 qubes (not unreasonable) is 618M of RAM, which is significant.  GVFS uses 31.8MB and evolution-data-server uses 18.8MB if started.  In `sys-net` evolution-data-server uses a more reasonable 2.1MB.

DemiMarie:
> But then, it needs proper issue for tracking and including in release notes.

Does this fall under QubesOS/qubes-issues#7028?

marmarek:
> Does this fall under [QubesOS/qubes-issues#7028](https://github.com/QubesOS/qubes-issues/issues/7028)?

It's related, but I think it should be separate issue (kinda sub-task), as #7028 is broader (not going to be solved by this PR alone).

marmarek:
This still doesn't build, on neither Debian nor Fedora

marmarek:
> * The default email client in Qubes is Thunderbird, not Evolution

Evolution is pre-installed in Debian, I think. Is evolution-data-server needed for Evolution to function, or is it just optional component? In the former case, can you make it enabled also when Evolution is installed (but possible to disable via qvm-service)?

DemiMarie:
> > * The default email client in Qubes is Thunderbird, not Evolution
> 
> Evolution is pre-installed in Debian, I think. Is evolution-data-server needed for Evolution to function, or is it just optional component? In the former case, can you make it enabled also when Evolution is installed (but possible to disable via qvm-service)?

On Fedora, Evolution at least manages to start and save a calendar event without evolution-data-server running.

marmarek:
> And an update to qvm-service man page (in core-admin-client repo).

Please add this too, then it will be good to merge.

marmarek:
Both commit message have faulty `qvm-service` call example - it's `--enable`, not `enable`. Alternatively, `qvm-service VMNAME tracker on`.

marmarek (approved):

PR #449, opened 2023-08-05
Remove outdated comment

andrewdavidwong:
Fixes QubesOS/qubes-issues#7237

### DIFF
--- a/passwordless-root/qubes.sudoers
+++ b/passwordless-root/qubes.sudoers
@@ -1,48 +1,4 @@
 Defaults !requiretty
 %qubes ALL=(ALL) ROLE=unconfined_r TYPE=unconfined_t NOPASSWD: ALL
 
-# WTF?! Have you lost your mind?!
-#
-# In Qubes VMs there is no point in isolating the root account from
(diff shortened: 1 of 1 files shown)


marmarek:
Maybe leave a link to https://www.qubes-os.org/doc/vm-sudo/ in its place?

andrewdavidwong:
I considered doing that but decided against it, because if https://www.qubes-os.org/doc/vm-sudo/ ever moves or changes (enough) again, this comment would have to be updated again (and look how long it's taken just for anyone to touch this comment). I'm guessing the motivation is explaining the rationale for a design decision, but it's not obvious that every design decision needs to contain its justification (or even a pointer to it) within the very same file. If the motivation is that people frequently ask about this when they see it, well, that's what FAQs are for.

Nonetheless, I don't have any particular investment in this (including the link) one way or the other, so if you'd prefer to include it, by all means, let's do that.

marmarek:
Can you make this PR against the "main" branch?

andrewdavidwong:
> Can you make this PR against the "main" branch?

Done: #452

PR #451, opened 2023-08-17
Fix qubes.TemplateSearch with repo_gpgcheck=1

DemiMarie:
qubes-core-agent-dom0-updates was missing a dependency on qubes-repo-templates, so the OpenPGP keys needed with repo_gpgcheck=1 weren't available to DNF.  Also add -y to the DNF command line so that it imports these keys without a prompt.

Fixes: QubesOS/qubes-issues#7414

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -125,6 +125,7 @@ Depends:
     dnf | yum-utils,
     curl,
     qubes-core-qrexec,
+    qubes-repo-templates,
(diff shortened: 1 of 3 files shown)


DemiMarie:
PipelineRetryFailed

marmarek (approved):
this is only a workaround for default repositories, but lets have at least this for now

PR #455, opened 2023-08-22
Disable multicast DNS and link-local multicast name resolution

DemiMarie:
These protocols increase the attack surface of a VM that is exposed to attackers that are network-adjacent to that VM, such as VMs that this VM provides network to or that provide network to the VM.  LLMNR is deprecated in favor of mDNS, which is provided by Avahi in Fedora. Qubes OS already disables Avahi by default, so disabling LLMNR and mDNS
in systemd-resolved is unlikely to break any existing workflows.

Fixes: QubesOS/qubes-issues#8444


marmarek:
First of all, any change in behavior needs an issue for tracking.
But then, the package fails to build (looks like `%_systemd_util_dir` isn't universally available).

marmarek (approved):

PR #456, opened 2023-09-16
qfile-agent: use a dirfd instead of getcwd()

rustybird:
Fixes QubesOS/qubes-issues#8516

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -3,7 +3,12 @@ checks:tests:
   - ci/codecov-wrapper
   before_script:
   - sudo dnf install -y python3-pip python3-gobject-base ShellCheck
+      'pkgconfig(icu-uc)'
   - pip3 install --quiet -r ci/requirements.txt
+  - git clone https://github.com/QubesOS/qubes-linux-utils ~/qubes-linux-utils
+  - make -C ~/qubes-linux-utils/qrexec-lib all
+  - sudo make -C ~/qubes-linux-utils/qrexec-lib install
+      LIBDIR=/usr/lib64 INCLUDEDIR=/usr/include
   script:
   - ./run-tests
   - shellcheck -e SC1117 $(grep -l '^#!/bin/\(ba\)\?sh' $(git ls-files))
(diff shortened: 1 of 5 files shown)


marmarek:
Thanks for working on this!
Can you try to also add/extend test for this case? We have very similar already at https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/tests/integ/vm_qrexec_gui.py#L605

rustybird:
Since [it turned out](https://github.com/QubesOS/qubes-issues/issues/8516#issuecomment-1722305933) that Demi had already fixed the data loss on a more general level (with that fix including a test) and it just wasn't backported to R4.1 yet:

I've added some test code here in this repo to test the tools in `qubes-rpc/` independently. The Filecopy test cases pipe `qfile-agent` into `qfile-unpacker` locally. `test_10_shadowed_by_mount` covers an example of the most essential failure mode related to the new bug, which is `qfile-agent` sending the wrong file *successfully*.

rustybird:
PipelineRetry

rustybird:
PipelineRetryFailed

rustybird:
Hmm in checks:tests it can't find the qubes-utils-libs and qubes-utils-devel packages for installation, and my tests are skipped.

Can/should I configure Qubes OS package repos somehow?

marmarek:
> Hmm in checks:tests it can't find the qubes-utils-libs and qubes-utils-devel packages for installation, and my tests are skipped.
> 
> Can/should I configure Qubes OS package repos somehow?

In other similar cases, we build dependencies in CI via a local git clone:
- https://github.com/QubesOS/qubes-core-qrexec/blob/main/.gitlab-ci.yml#L20-L22
- https://github.com/QubesOS/qubes-core-admin/blob/master/.gitlab-ci.yml#L14

This approach allows testing against `main` branch, even if no package is built, but also eases switching CI container to another distro (like bumping its version before proper packages are uploaded for given Fedora version).
If that would work here without too much complications, I'd prefer this approach. But if yuu run into some dependency issues (like needing to clone/build 3+ more repositories), then probably it isn't worth it, and its easier to add qubes repository then (accepting the downsides). Fortunately, core-agent-linux is the place where we have qubes repository definitions, so you don't need to copy them from elsewhere.

marmarek (approved):

PR #458, opened 2023-09-23
Provide a alternate qfile-unpacker that is not SELinux-confined

DemiMarie:
Using this in the Qubes executor will allow using the Qubes executor with SELinux enforcing.


DemiMarie:
PipelineRetryFailed

marmarek (approved):

PR #462, opened 2023-10-05
Add support for IPv6 Virtual DNS

1cho1ce:
Add support for IPv6 DNS together with this pull request:
https://github.com/QubesOS/qubes-core-admin/pull/558


marmarek:
Unfortunately this may not be as simple as this. All is fine if you really have IPv6 DNS in your network. But if you have IPv4 one (but otherwise have IPv6 connectivity), then queries sent to the "virtual DNS" address won't be redirected anywhere, most likely will be silently discarded somewhere along the way (either in sys-net, but more likely on some next router). This is bad, because from application point of view, this is no response and the application will wait for a timeout instead of immediately falling back to IPv4 DNS. And it is especially bad with infamous DNS fallback handling in Linux...
BTW, the same issue applies to the case of only IPv6 DNS (even without this PR), but it's a smaller problem because it's rather rare setup.

Some solution to this issue might be forwarding between IPv6 and IPv4 appropriately, for example using socat (`socat UDP-LISTEN:... UDP:...`). It's an ugly solution...

Or maybe there is some way to avoid long timeouts, like reject with some specific ICMP packet if no IPv6 DNS is present?

1cho1ce:
What if we add a reject rule for IPv6 DNS queries if it's a qube with:

1. IPv6 enabled
2. It's connected to outside network:
2.1. Either with network adapter passthrough (sys-net) (check if `/qubes-primary-dns` or `/qubes-ip` is `None`?)
2.2. Or it's a qube meant to encapsulate IPv6 traffic for only some subset of Qubes networking as described here for teredo example (check if `/qubes-ip6` is set and `/qubes-gateway6` is `None`?):
https://www.qubes-os.org/doc/networking/#ipv6
3. No real IPv6 DNS (not Qubes OS Virtual DNS address) found with `get_dns_resolved()`

Then IPv6 DNS queries will be rejected and there shouldn't be any timeouts or queries to Virtual DNS sent outside?

marmarek:
The second condition may not be needed in fact. Either qube has some IPv6 DNS it can forward traffic to (either "real" one outside, or virtual one, or via some VPN or such) or not. Specifically, if you have a VPN that is supposed to encapsulate all your traffic, it should adjust local DNS servers in that qube too - at which point the 3rd condition will handle this case.

marmarek:
But yes, something like this might work. Try it out and let me know :)

1cho1ce:
I've tested the reject rules and it works fine.
First one without reject or dnat DNS rules and second one with reject rules:
```
user@test:~$ time curl -o /dev/null -s https://debian.org

real	0m20.068s
user	0m0.009s
sys	0m0.018s
user@test:~$ time curl -o /dev/null -s https://debian.org

real	0m0.016s
user	0m0.009s
sys	0m0.006s
```

I had to change the way Qubes OS set DNS servers to qubes with these commits:
https://github.com/QubesOS/qubes-core-admin/pull/558/commits/98b184c183391694f860dea7051361e2f6595fcc
https://github.com/QubesOS/qubes-core-agent-linux/pull/462/commits/d659bfbdcaf65262c4678e9ff8f5705abb912096
Before it was setting DNS for qubes that provides network and don't have netvm (sys-net/sys-usb/etc) which I think is pointless (maybe I just don't understand where it would be used) and It interfered with implementation of DNS reject rules.
I'm not sure if this will break anything. I've tested that it didn't break basic things but maybe there are some things that could break.

Also I had to hardcode Qubes OS Virtual DNS addresses here:
https://github.com/QubesOS/qubes-core-agent-linux/pull/462/commits/0aaad6a9a1bb011be0b065289814d952f5c114fd#diff-1e7c156682261fecd8a773620728327528f3be00468837709d7ffeb8073cc602R29
Which is bad but I couldn't think of an easy way out.
If we just reject all DNS queries in dnat-dns chain with rules like this:
`tcp dport 53 reject with icmp type host-unreachable`
Then it'll reject all DNS queries that would come from qubes including:
- input queries to DNS server running in this qube
- forward queries to some specific "real" DNS which downstream qubes used instead of "virtual" DNS

marmarek:
> Also I had to hardcode Qubes OS Virtual DNS addresses here:

See my comment in the other PR.

Anyway, I think the general shape is getting there. But due to release stage of R4.2, it's too late to be included there, at least before final release. So, I'll merge it only after final R4.2 release.

marmarek:
> * VmNetworking_fedora-38-xfce: [test_000_simple_networking](https://openqa.qubes-os.org/tests/84894#step/VmNetworking_fedora-38-xfce/1) (failure)
>       `ValueError: too many values to unpack (expected 2)`

This looks broken (click the link for full traceback)

1cho1ce:
> > * VmNetworking_fedora-38-xfce: [test_000_simple_networking](https://openqa.qubes-os.org/tests/84894#step/VmNetworking_fedora-38-xfce/1) (failure)
> >   `ValueError: too many values to unpack (expected 2)`
> 
> This looks broken (click the link for full traceback)

Fixed and tested with systemd-resolved.

network/qubes-setup-dnat-to-ns
```diff
@@ -95,15 +104,60 @@ def install_firewall_rules(dns):
...
+        'type nat hook prerouting priority dstnat; policy accept;',
+        'jump custom-dnat-dns',
+    ]
+    for i, (vm_nameserver, dest) in enumerate(zip_longest(qubesdb_dns,
+        [dns_ip for dns_ip in dns if type(dns_ip) is IPv4Address])):
+        if i >= len(qubesdb_dns):
+            break
+        dns_ = str(dest)
+        if dest is None or (vm_nameserver == dest and
+                            qdb.read('/qubes-ip') is None):
```
SaswatPadhi, line 118:
```suggestion
                            qdb.read('/qubes-primary-dns') is None):
```

for consistency?

SaswatPadhi (commented):
Thanks @1cho1ce for working on this PR. I was trying to do something similar and came across this and your other PR.
Are you still working on this? I was going through the code and noticed some minor issues, so commenting below.

marmarek:
Yes, do you want to pick up this work?

marmarek:
Network integration tests are at https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/tests/integ/network_ipv6.py and https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/tests/integ/network.py. They include already basic tests like pinging by IP or name. A useful test would be to configure IPv6-only DNS and check that (but keep also tests for IPv4 DNS, and maybe add some for IPv4+IPv6 DNS too?). The `configure_netvm` function is responsible for setting up basic setup (including dnsmasq used as test DNS server), but individual tests can adjust the configuration later too.

I can take care of it later, but if anybody wants to do that earlier, feel free.

PR #465, opened 2023-10-08
qvm-copy(-to-vm.gnome): avoid 'find -files0-from' for bullseye

rustybird:


rustybird:
@marmarek should `repro:bookworm` maybe be marked `allowed to fail`?

marmarek:
> Should `repro:vm-bookworm` maybe be marked `allowed to fail`?

It's an issue only in this repo (https://github.com/QubesOS/qubes-issues/issues/8315) and I'm not sure how to do that with our configuration all included from one place...

rustybird:
Oh, I didn't realize that spamming `PipelineRetryFailed` will eventually work for `repro:vm-bookworm` even in this repo. I'll do that.

marmarek (approved):

rustybird:
@marmarek this still needs to be cherry-picked for R4.1

PR #473, opened 2023-11-07
qubes-run-terminal: add gnome-console support

zpc0:


marmarek:
PipelineRetry

app-menu/qubes-run-gnome-console
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+# Wrapper around gnome-console that runs it with --wait option
+# (if supoported).
+
+# Check if our gnome-console version supports --wait
+# (we can't just run it and check exit code, because if it works, it will
+# return the exit code of the child process)
+case "$(kgx --help)" in
+    *--wait\ *) exec kgx --wait "$@" ;;
```
marmarek, line 9: `kgx --help` says `TODO` next to this option...
Empirically, if that's the first kgx invocation, it waits for the command to complete (regardless of using --wait or not), but kgx is running already, then it exits immediately (also, regardless of using --wait option or not).

So, for the purpose not closing DispVM too early, running kgx directly should be fine (as that will indeed be the first kgx process), and this wrapper is unnecessary.
FWIW I checked it with gnome console 45.0. Did you need to use --wait option to keep the process waiting in some other version?

marmarek (changes requested):

app-menu/qubes-run-gnome-console
zpc0: Indeed, this wrapper is currently useless.
The purpose of this wrapper is to prevent problems like `gnome-console does not work with DispVM` when the `--wait` option is implemented in a future version.

If you want to keep program minimal, it is fine to remove this wrapper.

marmarek: I see, if they make then gnome-console _without_ `--wait` really not wait, then it will be an issue. Ok, lets keep the script.

marmarek (approved):

PR #474, opened 2023-11-21
Parse rc.local.d directory on boot

ben-grande:
Fixes https://github.com/QubesOS/qubes-issues/issues/8690

### DIFF
--- a/init/setup-rw.sh
+++ b/init/setup-rw.sh
@@ -31,7 +31,12 @@ if mountpoint -q /rw ; then
 # This script will be executed at every VM startup, you can place your own
 # custom commands here. This includes overriding some configuration in /etc,
 # starting services etc.
-
+#
+# Executable scripts located in /rw/config/rc.local.d with the extension
+# '.rc' are executed immediately before this rc.local.
+# Example:
+#  /rw/config/rc.local.d/custom.rc
+#
 # Example for overriding the whole CUPS configuration:
 #  rm -rf /etc/cups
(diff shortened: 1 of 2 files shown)


ben-grande:
Maybe I am late because OpenQA has already run but should I only parse files with extensions like `.conf`? Although it is a script, a extension might be useful to avoid parsing unwanted files...

- qubes-bind-dirs.d uses `.conf`
- qubes-firewall.d doesn't check extension

ben-grande:
On 23-12-01 05:56:57, Marek Marczykowski-Górecki wrote:
> > Although it is a script, a extension might be useful to avoid parsing unwanted files...
> 
> Generally it's a good idea, maybe just `.sh`? I find it weird to have arbitrary script with `.conf` (I know bind-dirs is such a case already, but it's supposed to configure specific variable, not run arbitrary commands, even if technically it's possible). Alternatively, require files to be executable, but that may be not enough to prevent running `.bak`, `.rpmsave` etc files...

I don't know of a suitable extension but it should not be ".sh" as it
doesn't need to be a ShellScript, it can be any type of executable file.

It is already requiring the file to be executable though, so an
extension is just a safeguard.

What do you think of ".rc"? As in:

- /rw/config/rc.local.d: any executable
- bashrc: ShellScript, configuration, function and commands
- vimrc: VimScript, configuration, function and commands

Although normally it is for shellscripts, let's take the meaning of
"runcom" or "run commands":

	https://unix.stackexchange.com/a/3469

-- 
Benjamin Grande

PR #480, opened 2024-01-07
Ensure XDG secrets portal is available

DemiMarie:
Sandboxed applications use this to store secrets.  Without it, Element cannot access its secrets and resets itself, causing user data to be lost.  The version of xdg-desktop-portal in Fedora 38 and below would use any portal implementation it could find, but the version in Fedora 39 requires that portal implementations be specified in a config file. Provide that config file.

Fixes: QubesOS/qubes-issues#8839

### DIFF
--- a/config-overrides/Makefile
+++ b/config-overrides/Makefile
@@ -19,3 +19,4 @@ install:
 		20_org.gnome.nautilus.qubes.gschema.override \
 		20_org.gnome.settings-daemon.plugins.updates.qubes.gschema.override \
 		20_org.mate.NotificationDaemon.qubes.gschema.override
+	install -D -m 0644 portals.conf $(DESTDIR)/usr/share/xdg-desktop-portal/portals.conf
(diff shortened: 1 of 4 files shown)


marmarek:
Missing `git add`? And also missing `$(DESTDIR)` in the makefile.
But also, how does it work on plain Fedora? I don't see any package providing this file there...

config-overrides/Makefile
```diff
@@ -19,3 +19,4 @@ install:
 		20_org.gnome.nautilus.qubes.gschema.override \
 		20_org.gnome.settings-daemon.plugins.updates.qubes.gschema.override \
 		20_org.mate.NotificationDaemon.qubes.gschema.override
+	install -D -m 0644 portals.conf /usr/share/xdg-desktop-portal/portals.conf
```
marmarek, line 22:
```suggestion
	install -D -m 0644 portals.conf $(DESTDIR)/usr/share/xdg-desktop-portal/portals.conf
```

marmarek (changes requested):

marmarek (approved):

PR #490, opened 2024-03-28
Recognize more SVG files

DemiMarie:
Without this change, GDK fails to recognize various icons as SVG files.
This causes Thunderbird to emit lots of warnings and, when the user
tries to attach an icon, crash on an assertion failure.

Fixes: https://github.com/QubesOS/qubes-issues/issues/9065


marmarek (changes requested):
Since you add `image/jpeg` type, add it to `qvm-view-in-dvm.desktop` too.

DemiMarie:
done

app-menu/qvm-open-in-dvm.desktop
```diff
@@ -3,8 +3,10 @@
 ###   (typing "xdg-settings set default-web-browser firefox.desktop" will put it back to normal)
 
 [Desktop Entry]
-Name=qvm-open-in-dvm
-Exec=qvm-open-in-dvm %u
+Version=1.0
+Name=QubesOS View In DisposableVM
+Exec=/usr/bin/qvm-open-in-dvm --view-only -- %u
```
marmarek, line 8: You changed "edit in disposable" to "view in disposable", that shouldn't be the case here.

marmarek (changes requested):

marmarek (approved):

PR #491, opened 2024-03-28
Enable Arch repositories

DemiMarie:
I'm not sure whether to have the testing repository uncommented or not, given that Arch is a rolling release distribution.  The duplicated Server= line is a workaround for a libalpm bug that causes Pacman to segfault.


marmarek:
It's already done in https://github.com/QubesOS/qubes-meta-packages/tree/main/archlinux

DemiMarie:
Does that mean this file should just be deleted?  I don’t think it serves any purpose.

marmarek:
It can be deleted, yes

PR #497, opened 2024-05-09
Allow unsafe characters if invoked as qubes.UnsafeFileCopy

DemiMarie:
This uses QREXEC_SERVICE_FULL_NAME to detect what the service was invoked as.  Non-empty arguments are reserved for future use.
    
This requires QubesOS/qubes-linux-utils#113.
    
Part of QubesOS/qubes-issues#8332

### DIFF
--- a/qubes-rpc/qfile-unpacker.c
+++ b/qubes-rpc/qfile-unpacker.c
@@ -113,7 +113,7 @@ int main(int argc, char ** argv)
             if (asprintf(&procdir_path, "/proc/%d/fd", getpid()) < 0) {
                 gui_fatal("Error allocating memory");
             }
-            procfs_fd = open(procdir_path, O_DIRECTORY | O_RDONLY);
+            procfs_fd = open(procdir_path, O_DIRECTORY | O_RDONLY | O_NOCTTY | O_CLOEXEC);
             if (procfs_fd < 0)
                 perror("Failed to open /proc");
             else
--- a/qubes-rpc/qfile-unpacker.c
+++ b/qubes-rpc/qfile-unpacker.c
@@ -11,8 +11,13 @@
 #include <string.h>
 #include <unistd.h>
 #include <sys/fsuid.h>
-#include <gui-fatal.h>
 #include <errno.h>
+#include <assert.h>
+#include <limits.h>
+#include <getopt.h>
+#include <err.h>
+
+#include <gui-fatal.h>
 #include <libqubes-rpc-filecopy.h>
 
 #define INCOMING_DIR_NAME "QubesIncoming"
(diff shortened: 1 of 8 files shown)


qubes-rpc/qfile-unpacker.c
```diff
@@ -44,31 +45,61 @@ int main(int argc, char ** argv)
...
         incoming_dir = argv[2];
     } else {
-        uid = getuid();
         home_dir = prepare_creds_return_dir(uid);
         remote_domain = getenv("QREXEC_REMOTE_DOMAIN");
         if (!remote_domain) {
             gui_fatal("Cannot get remote domain name");
         }
 
+        const char *service = getenv("QREXEC_SERVICE_FULL_NAME");
```
marmarek, line 73: Please don't, use a command line switch instead (the `qfile-unpacker uid dir` invocation must still work, but it can easily be supported as `qfile-unpacker [options] [uid dir]`). It's easy since `/etc/qubes-rpc/qubes.Filecopy` is a script.
See task description at https://github.com/QubesOS/qubes-issues/issues/8332

marmarek (changes requested):

qubes-rpc/qfile-unpacker.c
DemiMarie: I’ll have two cases:

1. `argv[1]` starts with a digit: uses current command line parsing code and does not support new features.  Possibly emits a deprecation warning.
2. `argv[1]` does _not_ start with a digit: uses `getopt_long(3)` to parse options and supports all features.

marmarek: Do you really need two cases? `getopt_long` should be good in any case, and have uid, dir as non-option arguments left after the call (I don't think you need to care about dir name conflicting with an option, it isn't attacker controlled at all)

marmarek: Ok, with one exception:

> The new calling convention will require `--` before non-option arguments, so there is no ambiguity.

No. It can support `--`, but it should not _require_ it.

DemiMarie: People know that they need to include `--` before non-option arguments that might begin with `-`, but in this case, a non-option argument that begins with an ASCII digit could _also_ cause problems.  Requiring `--` means that people will _always_ include it, and therefore get correct behavior every time.

marmarek: Still, do not require `--`. If somebody wants to include `--`, they can.

marmarek (changes requested):

marmarek (changes requested):
This PR is missing actually introducing `qubes.UnsafeFilecopy` service. And also changing `qvm-copy` to choose service based on files to be copied (see https://github.com/QubesOS/qubes-issues/issues/8332#issuecomment-1912403043)

marmarek:
This fails to build on Debian:
```
12:57:49,692 [executor:docker:0e16ac29c7ff] output: cc -g -pie -Wl,-z,relro '-Wl,-rpath,$ORIGIN/../../$LIB'  vm-file-editor.o  -lqubes-rpc-filecopy -lqubes-pure -o vm-file-editor
12:57:49,697 [executor:docker:0e16ac29c7ff] output: /usr/bin/ld: cannot find -lqubes-pure: No such file or directory
```
Likely a missing Build-Depends:

marmarek:
In fact it fails to build on Fedora too, due to a different reason:
```
18:50:55,505 [executor:docker:5697546c802e] output: DEBUG:     Installed (but unpackaged) file(s) found:
18:50:55,505 [executor:docker:5697546c802e] output: DEBUG:    /usr/lib/debug/usr/lib/qubes/qubes-fs-tree-check-4.3.0-1.2.fc40.x86_64.debug
18:50:55,506 [executor:docker:5697546c802e] output: DEBUG:    /usr/lib/qubes/qubes-fs-tree-check
```

DemiMarie:
Build failed because QubesOS/qubes-linux-utils#113 is not part of the latest released package.

marmarek:
qvm-copy tests fail (see updated openqa report above), example failure:
```
# test_100_qrexec_filecopy
# failure: 

# timestamp 2024-05-29T03:57:36.815652
  File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 612, in test_100_qrexec_filecopy
    self.loop.run_until_complete(
  File "/usr/lib64/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
subprocess.CalledProcessError: Command 'qvm-copy-to-vm test-inst-vm2 /tmp/passwd' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

  File "/usr/lib64/python3.11/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/qubes/tests/integ/vm_qrexec_gui.py", line 617, in test_100_qrexec_filecopy
    self.fail('qvm-copy-to-vm failed: {}'.format(e.stderr))
AssertionError: qvm-copy-to-vm failed: b'qubes-fs-tree-check: Argument 5 is an absolute path\n'
```

DemiMarie:
This is due to an outdated check from when I misunderstood how `qfile-agent` selects the filenames to pass to the remote side.  I thought it used the full path, but it only uses the basename.

marmarek:
Manual test passed.
But it may need adjustment for https://github.com/QubesOS/qubes-issues/issues/8332#issuecomment-2135628595

DemiMarie:
I will update the code to use the service argument.

marmarek (approved):

PR #500, opened 2024-05-16
pacman hook to notify dom0 about successful upgrade

alimirjamali:
Resetting `updates-available` feature of Archlinux template after successful upgrades

Reference to issue: Fixes QubesOS/qubes-issues#9233

### DIFF
--- a/archlinux/PKGBUILD.qubes-post-upgrade.hook
+++ b/archlinux/PKGBUILD.qubes-post-upgrade.hook
@@ -1,0 +1,10 @@
+[Trigger]
+Operation = Upgrade
+Type = Path
+Target = *
+
+[Action]
+Description = Notifying dom0 about successful upgrade...
+Depends = qubes-vm-qrexec
+When = PostTransaction
+Exec = /usr/lib/qubes/qrexec-client-vm dom0 qubes.NotifyUpdates /bin/sh -c 'echo 0'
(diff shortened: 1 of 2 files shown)


lubellier (approved):

archlinux/PKGBUILD.qubes-post-upgrade.hook
```diff
@@ -0,0 +1,10 @@
+[Trigger]
+Operation = Upgrade
+Type = Path
+Target = *
+
+[Action]
+Description = Notifying dom0 about successful upgrade...
+Depends = qubes-vm-qrexec
+When = PostTransaction
+Exec = /usr/lib/qubes/qrexec-client-vm dom0 qubes.NotifyUpdates /bin/sh -c 'echo 0'
```
marmarek, line 10: Shouldn't this check if actually all updates got applied (like, if it wasn't update of a single package out of several updates available)?
For example, if I do `pacman -S xterm` (having xterm installed but at older version), it clears the updates available flag, even if several other updates are available.
Calling `/usr/lib/qubes/upgrades-status-notify` instead of this command should be enough.

marmarek (changes requested):

alimirjamali:
@marmarek Running `upgrade-status-notify` as a post-upgrade hook for Archlinux does not work in this case. I have tried it and discussed it on [forum](https://forum.qubes-os.org/t/improvements-for-archlinux-and-debian-template-updates/26395/4) last week. `upgrades-installed-check` scripts uses `checkupdates(8)` for Archlinux and it does not reset during post-upgrade hooks. Only after them.

I also considered a single package installation/upgrade. This hook does not run during new package installations. But it will run if someone re-installs an existing package or only upgrade one or few of them. I considered it an special case where usually Archlinux users know what they are doing and partial upgrade should be intentional. So it might be better to clear updates-available feature in such cases.

marmarek:
I see... in this case, I guess it's a reasonable trade-off, given partial upgrades are IIUC discouraged in Archlinux anyway. If some updates remains to be installed, next updates check will notice it.

marmarek (approved):

PR #502, opened 2024-05-25
Always pass -y to DNF when downloading dom0 updates

DemiMarie:
When downloading updates for dom0, DNF's stdin is not connected to any terminal, so it makes no sense for DNF to ever prompt for interactive confirmation.  By default, DNF would prompt for confirmation before importing a metadata signing key, causing breakage.

This change fixes `qubes-dom0-update --check-only` using sys-whonix as UpdateVM.

Fixes: QubesOS/qubes-issues#9212

### DIFF
--- a/package-managers/qubes-download-dom0-updates.sh
+++ b/package-managers/qubes-download-dom0-updates.sh
@@ -60,9 +60,9 @@ if [ -z "$YUM_ACTION" ]; then
 fi
 
 if type dnf >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


marmarek (changes requested):
Now the `-y` is duplicated in a bunch of commands below.

DemiMarie:
> Now the `-y` is duplicated in a bunch of commands below.

harmless, but I’ll fix it.

marmarek (approved):

PR #504, opened 2024-06-06
Add os-release ID_LIKE to features

ben-grande:
Very useful to apply configuration based on the base distribution that the OS was created upon, instead of adding an extensive list of supported distributions of Debian.

### DIFF
--- a/qubes-rpc/post-install.d/10-qubes-core-agent-features.sh
+++ b/qubes-rpc/post-install.d/10-qubes-core-agent-features.sh
@@ -1,45 +1,89 @@
 #!/bin/sh
+#
+# The Qubes OS Project, http://www.qubes-os.org
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+#
+# URL: https://github.com/dylanaraps/pfetch
+# License: MIT
+# Copyright (c) 2016-2019 Dylan Araps
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to
+# deal in the Software without restriction, including without limitation the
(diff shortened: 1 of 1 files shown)


qubes-rpc/post-install.d/10-qubes-core-agent-features.sh
```diff
@@ -16,19 +15,18 @@ if [ -r /etc/os-release ]; then
...
     qvm-features-request os-distribution="$distro"
 
-    version=$(grep ^VERSION_ID= /etc/os-release)
-    version=${version#VERSION_ID=}
-    version=${version#\"}
-    version=${version%\"}
+    distro_like=$(awk -F '=' '/^ID_LIKE=/{ print $2 }' /etc/os-release)
+    qvm-features-request os-distribution-like="${distro_like}"
+
+    version=$(awk -F '[="]' '/^VERSION_ID=/{ print $3 }' /etc/os-release)
```
marmarek, line 21: This needs a bit more care - VERSION_ID value sometimes it quoted but sometimes not

qubes-rpc/post-install.d/10-qubes-core-agent-features.sh
```diff
@@ -16,19 +15,18 @@ if [ -r /etc/os-release ]; then
     fi
     qvm-features-request os-distribution="$distro"
 
-    version=$(grep ^VERSION_ID= /etc/os-release)
-    version=${version#VERSION_ID=}
-    version=${version#\"}
-    version=${version%\"}
+    distro_like=$(awk -F '=' '/^ID_LIKE=/{ print $2 }' /etc/os-release)
+    qvm-features-request os-distribution-like="${distro_like}"
```
marmarek, line 19: This needs also the backend side, see https://github.com/QubesOS/qubes-core-admin/pull/578 and later https://github.com/QubesOS/qubes-core-admin/pull/585.

What is the "extensive list of supported distributions of Debian" what you can find on Qubes?

marmarek (changes requested):

qubes-rpc/post-install.d/10-qubes-core-agent-features.sh
ben-grande: Last commits deals with this now.

qubes-rpc/post-install.d/10-qubes-core-agent-features.sh
ben-grande: > This needs also the backend side, see [QubesOS/qubes-core-admin#578](https://github.com/QubesOS/qubes-core-admin/pull/578) and later [QubesOS/qubes-core-admin#585](https://github.com/QubesOS/qubes-core-admin/pull/585).

Will take care of this soon.

> What is the "extensive list of supported distributions of Debian" what you can find on Qubes?

That I can find on Qubes by default is only `kali` and `whonix`. That someone can install from a 3rd party template, that is higher.

ben-grande:
I didn't think grep, awk, sed, cut, head where a good use of resources. Got some code from `pfetch` that uses built-in file reading and string replacement. Also deals with quotes.

marmarek (approved):

PR #505, opened 2024-06-19
Fixing issue 9011: DNS leakage when only one DNS server is set

alimirjamali:
Fixes https://github.com/QubesOS/qubes-issues/issues/9011
Details in Github comments and forum link

### DIFF
--- a/network/qubes-setup-dnat-to-ns
+++ b/network/qubes-setup-dnat-to-ns
@@ -53,8 +53,6 @@ def get_dns_resolved():
     try:
         resolve1 = bus.get_object('org.freedesktop.resolve1',
                                   '/org/freedesktop/resolve1')
-        resolve1_proxy = dbus.Interface(resolve1,
-                                        dbus_interface='org.freedesktop.resolve1')
         dns = resolve1.Get('org.freedesktop.resolve1.Manager',
                            'DNS',
                            dbus_interface='org.freedesktop.DBus.Properties')
@@ -71,7 +69,10 @@ def get_dns_resolved():
     dns.sort(key=lambda x: x[0] != 0)
     # Only keep IPv4 entries. systemd-resolved is trusted to return valid
     # addresses.
-    return [IPv4Address(bytes(addr)) for _g, family, addr in dns if family == 2]
+    # ToDo: We only need abridged IPv4 DNS entries for ifindex == 0.
+    # to ensure static DNS of disconnected network interfaces are not added.
+    return [IPv4Address(bytes(addr)) for ifindex, family, addr in dns
+            if family == 2]
 
 def install_firewall_rules(dns):
     qdb = qubesdb.QubesDB()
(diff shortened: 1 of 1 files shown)


marmarek:
> * [unstable] suspend: [Failed](https://openqa.qubes-os.org/tests/103627#step/suspend/33) (test died + timed out)
>       `# Test died: command 'qvm-run -p sys-firewall "curl https://www.qub...`

This looks to be broken by this PR. After suspend `dnat-dns` in sys-net looks like this:
```
	chain dnat-dns {
		type nat hook prerouting priority dstnat; policy accept;
		ip daddr 10.139.1.1 udp dport 53 drop
		ip daddr 10.139.1.1 tcp dport 53 drop
		ip daddr 10.139.1.2 udp dport 53 drop
		ip daddr 10.139.1.2 tcp dport 53 drop
	}
```
Yet, DNS in sys-net is set:
```
[root@dom0 ~]# qvm-run -pu root sys-net resolvectl
Global
         Protocols: -LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
  resolv.conf mode: foreign
Current DNS Server: 172.16.1.1
       DNS Servers: 172.16.1.1
        DNS Domain: testnet

Link 4 (vif5.0)
    Current Scopes: none
         Protocols: -DefaultRoute -LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported

Link 5 (ens6)
    Current Scopes: DNS
         Protocols: +DefaultRoute -LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
Current DNS Server: 172.16.1.1
       DNS Servers: 172.16.1.1
        DNS Domain: testnet

Link 6 (wls7)
    Current Scopes: none
         Protocols: -DefaultRoute -LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
[root@dom0 ~]# qvm-run -pu root sys-net cat /etc/resolv.conf
# Generated by NetworkManager
search testnet
nameserver 172.16.1.1
```

My guess is the extra `ifindex == 0` condition is to blame.

alimirjamali:
> My guess is the extra `ifindex == 0` condition is to blame.

The ifindex 0 is to fetch the final DNS construction based on the of DNS values of all network interfaces according to their static or dynamic DHCP settings as well as their connection status. That is what ends in the systemd-resolved generated `/etc/resolv.conf`. Freedesktop [specification](https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.resolve1.html) does not explicitly mention this. I found about it via manual query of dbus and comparing the returned values with the actual `/etc/resolv.conf` content. So there should be another bug somewhere else that prevents system-resolved to reconstruct network settings after resume from suspend. 

But for the time, removing `ifindex == 0` condition will prevent the solution from breaking connectivity for most users. I will  just amend the pull-request and force push it in few minutes. But I will leave the comments for further investigation.

I suggest a new Unit Test for further investigation. Put the system in sleep mode, Change upstream network parameters (DNS, Gateway, DHCP lease range, ...) , Then resume from sleep. The sys-net of the virtual machine should be able to connect to new upstream and apply the settings to have connectivity.

ben-grande:
@alimirjamali can you please reword the commit message? Remove `Fixing issue 9011` from the subject and put in the end of the body `Fixes: LINKTOISSUE`. I am not a Qubes developer to enforce this decision, but I'd be glad if this standard was followed for all commits. Qubes [commit message guidelines](https://www.qubes-os.org/doc/coding-style/#commit-message-guidelines) links to [another guide that shows an example with "Resolves: #123](https://cbea.ms/git-commit/). Note that full link to URL is preferred rather than a `#N` as Qubes has multiple repositories and sometimes  (rare) issues are opened in individual repositories rather than `qubes-issues`.

alimirjamali:
> can you please reword the commit message?

Done. Is it OK now?

ben-grande:
Yes, thank you.

marmarek (approved):

PR #511, opened 2024-07-16
Manual for qvm-move(1)

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/8023

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -13,6 +13,8 @@ help:
 install: manpages
 	mkdir -p $(DESTDIR)/usr/share/man/man1
 	cp $(VM_DOCS) $(DESTDIR)/usr/share/man/man1/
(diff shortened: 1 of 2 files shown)


marmarek:
Since you describe both qvm-move and qvm-move-to-vm, IMO there should be also a symlink so `man qvm-move-to-vm` works too. `doc/Makefile` should be a good place to make it. And while at it, add it for qvm-copy-to-vm too?

alimirjamali:
> Since you describe both qvm-move and qvm-move-to-vm, IMO there should be also a symlink so `man qvm-move-to-vm` works too. `doc/Makefile` should be a good place to make it. And while at it, add it for qvm-copy-to-vm too?

Done

alimirjamali:
For some reason, I can not open the Codecov reports. Please let me know if anything is broken.

marmarek (approved):

PR #513, opened 2024-07-18
Accept only files for qvm-open-in-(d)vm

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/9360

### DIFF
--- a/qubes-rpc/qvm-open-in-vm
+++ b/qubes-rpc/qvm-open-in-vm
@@ -64,6 +64,18 @@ if [ -z "$target" ] || [ -z "$filename" ]; then
     usage
 fi
 
+if [ -d "$filename" ]; then
+    printf 'Error! "%s" is a directory.\n' "$filename"
+    usage
+elif [ ! -f "$filename" ]; then
+    expr "@$filename" : '@[A-Za-z][A-Za-z0-9+]*://'>/dev/null
+    # shellcheck disable=SC2181
+    if [ $? -ne 0 ]; then
+        printf 'Error! "%s" is not a valid file or URL.\n' "$filename"
+        usage
+    fi
+fi
+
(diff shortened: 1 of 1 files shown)


alimirjamali:
Accepting URLs should be added to the conditions. Fix will come in few minutes...

alimirjamali:
According to tests:
```
In qubes-rpc/qvm-open-in-vm line 73:
elif [[ ! "$filename" =~ (https?|ftp|mailto|irc|rsync)://.*$ ]] && \
     ^-- SC3010 (warning): In POSIX sh, [[ ]] is undefined.
For more information:
  https://www.shellcheck.net/wiki/SC3010 -- In POSIX sh, [[ ]] is undefined.
```
Instead of upgrading the script to bash to allow regex, I believe it would be better to revert to the conventional `expr` URL evaluation which is already used in the script. Handling malformed or invalid URLs was not the main subject of this PR from the beginning.

qubes-rpc/qvm-open-in-vm
```diff
@@ -64,6 +64,21 @@ if [ -z "$target" ] || [ -z "$filename" ]; then
     usage
 fi
 
+if [ -d "$filename" ]; then
+    printf 'Error! "%s" is a directory.\n' "$filename"
+    usage
+elif [ -L "$filename" ]; then
```
marmarek, line 70: Why reject symlinks? They open just fine (but when saving modified file, symlink is replaced with a copy of the file, and the actual target file remains unchanged).

alimirjamali: > Why reject symlinks? They open just fine (but when saving modified file, symlink is replaced with a copy of the file, and the actual target file remains unchanged).

I was not aware of that. Symlinks are now allowed in the amended PR.

marmarek (approved):

PR #514, opened 2024-07-22
script to cleanup empty QubesIncoming dir

kennethrrosen:
Resolves QubesOS/qubes-issues/issues/8307


kennethrrosen:
As noted by @alimirjamali we could be more forgiving, and leave untouched if the directory is not empty:

```
if [ -d /home/user/QubesIncomming ]; then
    find /home/user/QubesIncoming -type f -empty -delete
    find /home/user/QubesIncoming -type d -empty -delete
fi
```

PR #515, opened 2024-07-22
 Misc-post-stop: remove unused QubesIncoming

kennethrrosen:
resolves QubesOS/qubes-issues#8307

### DIFF
--- a/vm-systemd/misc-post-stop.sh
+++ b/vm-systemd/misc-post-stop.sh
@@ -1,5 +1,10 @@
 #!/bin/sh
 
-# Save default applications for DispVM
+# cleanup QubesIncoming/
+if [ -d /home/user/QubesIncomming ]; then
+    find /home/user/QubesIncoming -mindepth 1 -type f -empty -delete
+    find /home/user/QubesIncoming -mindepth 1 -type d -empty -delete
+fi
 
+# Save default applications for DispVM
 exit 0
--- a/vm-systemd/misc-post-stop.sh
+++ b/vm-systemd/misc-post-stop.sh
@@ -1,9 +1,17 @@
 #!/bin/sh
 
+USERNAME="$(qubesdb-read /default-user || echo 'user')"
+INCOMING_DIR="/home/$USERNAME/QubesIncoming"
+
+# Check for opt-out mechanism
+if [ -f /run/qubes-service/no-qubesincoming-cleanup ]; then
+    exit 0
+fi
+
 # cleanup QubesIncoming/
-if [ -d /home/user/QubesIncomming ]; then
-    find /home/user/QubesIncoming -mindepth 1 -type f -empty -delete
-    find /home/user/QubesIncoming -mindepth 1 -type d -empty -delete
+if [ -d "$INCOMING_DIR" ]; then
+    find "$INCOMING_DIR" -mindepth 1 -maxdepth 1 -type f -empty -delete
+    find "$INCOMING_DIR" -mindepth 1 -maxdepth 1 -type d -empty -delete
 fi
 
 # Save default applications for DispVM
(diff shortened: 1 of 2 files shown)


alimirjamali (approved):

kennethrrosen:
> Can you please cleanup the git history here - include just the last commit?

@marmarek done

kennethrrosen:
I've made a minor change based on a discussion in the forum: https://forum.qubes-os.org/t/github-issue-8307-automatically-clean-up-qubesincoming-directory-if-empty/27798/14

vm-systemd/misc-post-stop.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
 
-# Save default applications for DispVM
+# cleanup QubesIncoming/
+if [ -d /home/user/QubesIncomming ]; then
```
marmarek, line 4:
```suggestion
if [ -d /home/user/QubesIncoming ]; then
```

And also don't hardcode `user`. See for example `qubes-rpc/qubes.WaitForSession` how to get the user name.

vm-systemd/misc-post-stop.sh
```diff
@@ -1,5 +1,10 @@
...
+    find /home/user/QubesIncoming -mindepth 1 -type f -empty -delete
+    find /home/user/QubesIncoming -mindepth 1 -type d -empty -delete
```
marmarek, line 6: Save the path (with the above adjustments) in a variable to not repeat it each time?

marmarek (changes requested):
Besides comments below, typo in the commit message (`reaolves`)

rustybird:
Sorry for chiming in late. I had assumed that this functionality is intended to be optional (opt-in or opt-out) but it looks like it's always-on. Isn't the effect a bit intense with that in mind?

Deleting `~/QubesIncoming/source-vm/` if it's completely empty seems relatively uncontroversial to me. (Although the reasoning for *not* deleting an empty `~/QubesIncoming/` - vanishing file manager bookmarks - could apply here too for some people's workflows.)

But recursively deleting empty directories at arbitrary depths, and especially empty files, can "randomly" mangle complex directory trees. Like, occasionally I'll move a downloaded tarball or similar (e.g. a software package) from a DispVM to another VM and then leave all the extracted contents lying around for a while, right there inside of `~/QubesIncoming/source-vm/`. These can contain meaningful empty directories or empty marker files.

kennethrrosen:
@marmarek resolved. gave this some time to let others chime in on the change.

alimirjamali (approved):

alimirjamali (approved):

marmarek:
> deleting empty directories at arbitrary depths, and especially empty files, can "randomly" mangle complex directory trees. Like, occasionally I'll move a downloaded tarball or similar (e.g. a software package) from a DispVM to another VM and then leave all the extracted contents lying around for a while, right there inside of `~/QubesIncoming/source-vm/`. These can contain meaningful empty directories or empty marker files.

This is a valid concern. I'd say removing empty files/dirs directly in `~/QubesIncoming/source-vm/` (and empty `source-vm` itself) should be okay, but going deeper is problematic. And also, an opt-out mechanism would be useful, for example check for `no-qubesincoming-cleanup` qvm-service (skip if `/run/qubes-service/no-qubesincoming-cleanup` is present)

kennethrrosen:
> and also, an opt-out mechanism would be useful, for example check for `no-qubesincoming-cleanup` qvm-service (skip if `/run/qubes-service/no-qubesincoming-cleanup` is present)

@marmarek I can add
```
# Check for opt-out mechanism
if [ -f /run/qubes-service/no-qubesincoming-cleanup ]; then
    exit 0
fi
```
 And maybe even a note at the top of the script to instruct users on adding their own qubes service until it's added by default in the qube manager GUI? Or I could make a PR for that with some guidance from @alimirjamali

marmarek:
There is a qvm-service discovery mechanism so it shows up in a dropdown in the qube settings (services tab). See `qubes-rpc/post-install.d/10-qubes-core-agent-features.sh`, you can add there `qvm-features-request supported.service.no-qubesincoming-cleanup=1`

alimirjamali:
> but going deeper is problematic

This is easily resolvable @kennethrrosen. You could add `--maxdepth 1` to find commands to resolve this.

marmarek (approved):

PR #516, opened 2024-07-24
 setup-ip: Invoke 'ip' via $PATH 

HW42:
Debian recently [1] removed the symlink in /sbin which breaks our setup-ip script.

Instead of hardcoding the new path, rely on normal $PATH resolution. Saving a few file existence checks isn't worth such breakage, I think (or if that really matters here, shell is the wrong tool in the first place).

[1]: https://salsa.debian.org/kernel-team/iproute2/-/commit/c4bb148dd4ed0601ca32ee8a458007d0c348d6c3

---

Assuming you agree with my reasoning above, one could argue to apply this change to all invocations of programs normally in the $PATH. Given that usage was already inconsistent I for now kept the change minimal instead.

The fix (this version or an update to /bin/ip) should be backported to 4.2.

### DIFF
--- a/network/setup-ip
+++ b/network/setup-ip
@@ -6,11 +6,11 @@
 set -ueo pipefail
 
 add_link_route () {
-    /sbin/ip -- route replace to unicast "$1" dev "$2" scope link
+    ip -- route replace to unicast "$1" dev "$2" scope link
 }
 
 add_default_route () {
-    /sbin/ip -- route replace to unicast default via "$1" dev "$2" onlink
+    ip -- route replace to unicast default via "$1" dev "$2" onlink
 }
 
 readonly netvm_mac=fe:ff:ff:ff:ff:ff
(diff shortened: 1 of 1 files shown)


marmarek:
For tools normally in /bin (or similar), I would agree that calling by just name is okay. But for tools in /sbin (or similar), better not, ironically exactly because of Debian - as (/usr)/sbin is not in default `$PATH`.  And this happens to be an issue also in various service contexts (like user systemd units), but also for scripts/tools used as normal users (for example `modinfo` really doesn't need to be called as root).

marmarek:
Yes, you can update commit message. I've recently added better reporting which commit was used by openqa, so there is no reason to avoid (especially non-functional) changes when label is already added (at worst, the test will need to be repeated, but if some changes needing re-tested are necessary, that would be the case anyway).

marmarek:
And BTW, on Fedora there is only `/usr/sbin/ip`, but no `/usr/bin/ip`, so with the symlink in Debian gone, there is no full path that would work on both anymore...

marmarek (approved):

PR #520, opened 2024-08-16
Allow recent:/// in Nautilus for Copy/Move/Open

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/8589

### DIFF
--- a/qubes-rpc/nautilus/qvm_copy_nautilus.py
+++ b/qubes-rpc/nautilus/qvm_copy_nautilus.py
@@ -1,5 +1,5 @@
-from gi.repository import Nautilus, GObject, GLib
-
+import os.path
+from gi.repository import Nautilus, GObject, GLib, Gio
 
 class CopyToAppvmItemExtension(GObject.GObject, Nautilus.MenuProvider):
     '''Copy file(s) to AppVM.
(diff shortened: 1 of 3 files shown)


qubes-rpc/nautilus/qvm_copy_nautilus.py
```diff
@@ -28,10 +28,30 @@ def get_file_items(self, *args):
...
+            elif file_uri.startswith('recent:///'):
+                try:
+                    file_info = file_location.query_info(
+                            Gio.FILE_ATTRIBUTE_STANDARD_TARGET_URI, 0, None)
+                    target_uri = file_info.get_attribute_string(
+                            Gio.FILE_ATTRIBUTE_STANDARD_TARGET_URI)
+                    if target_uri.startswith('file://'):
+                        paths.append(target_uri[7:])
+                except GLib.GError:
+                    # TODO: Decide what to do if the recent item does not exist
```
marmarta, line 48: I think that we should at least notify the user that something went wrong in those cases - and, hopefully, what went wrong. Like a "the file was moved" or something error?

marmarta (changes requested):

qubes-rpc/nautilus/qvm_copy_nautilus.py
alimirjamali: There is also another approach. To do it how it is [currently done](https://github.com/QubesOS/qubes-app-linux-img-converter/blob/main/qvm_convert_img_nautilus.py#L31-L34) in `Convert to Trusted Img` and `Convert to Trusted PDF` code.

Just hide the options if the URL is neither of  `file:///` or `recent:///`. I think it is better to remove the option from right-click menu for Network Shares for the time.

Please let me know if it is OK with you.

marmarta: Oh, I like this idea. Yeah, it's safer, I think.

alimirjamali: @marmarta It is done.

alimirjamali:
Additional development notes for reference:

Conversion of `recent:///UUID` URIs to the relevant `file:///...` or the equivalent URIs is not well documented. Looking at  [glib/gio/gio-tool-tree.c](https://github.com/GNOME/glib/blob/main/gio/gio-tool-tree.c#L56-L57) (`gio tree recent://` command) is the proper way to reverse engineer it to Python.

marmarta (approved):

PR #521, opened 2024-08-22
Support dnf5 for qubes-download-dom0-update.sh

alimirjamali:
dnf5 does not print "--downloadonly" help with -h by default

For https://github.com/QubesOS/qubes-issues/issues/9244

### DIFF
--- a/package-managers/qubes-download-dom0-updates.sh
+++ b/package-managers/qubes-download-dom0-updates.sh
@@ -144,53 +144,14 @@ if [ ${#PKGLIST[@]} -eq 0 ] && [ "$CHECK_ONLY" = "1" ]; then
     fi
 fi
 
-# now, we will download something
-UPDATE_COMMAND=(fakeroot $UPDATE_CMD $UPDATE_ACTION "${UPDATE_ARGUMENTS[@]}")
-"$UPDATE_CMD" "$UPDATE_ACTION" --help | grep -q downloadonly && UPDATE_COMMAND+=(--downloadonly)
-# check for --downloadonly option - if not supported (Debian), fallback to
-# yumdownloader
-if ! "$UPDATE_CMD" --help | grep -q downloadonly && ! "$UPDATE_CMD" --version | grep -q dnf5; then
-    if dpkg --compare-versions \
-            "$(dpkg-query --show --showformat='${version}' rpm)" gt 4.14; then
-        SIGNATURE_REGEX="^[A-Za-z0-9._+-/]{1,128}\.rpm: digests signatures OK$"
-    else
-        SIGNATURE_REGEX="^[A-Za-z0-9._+-/]{1,128}\.rpm: [a-z0-9() ]* (pgp|gpg) [a-z0-9 ]* OK$"
-    fi
+# now, we will download something (or perform search, list or other tasks)
+UPDATE_COMMAND=(fakeroot "$UPDATE_CMD" "$UPDATE_ACTION" "${UPDATE_ARGUMENTS[@]}")
 
-    # setup environment for yumdownloader to be happy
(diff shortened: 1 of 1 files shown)


alimirjamali:
Using `--version` to detect dnf5 in Debian scenario

alimirjamali:
Reason for amendments:

In dnf5:

1. `check-update` is renamed to `check-upgrade`
2. The order of command & argument does matter very much.

alimirjamali:
PipelineRetry

package-managers/qubes-download-dom0-updates.sh
```diff
@@ -129,10 +145,11 @@ if [ ${#PKGLIST[@]} -eq 0 ] && [ "$CHECK_ONLY" = "1" ]; then
 fi
 
 # now, we will download something
-YUM_COMMAND=(fakeroot "${YUM[@]}" "$YUM_ACTION" --downloadonly)
+UPDATE_COMMAND=(fakeroot $UPDATE_CMD $UPDATE_ACTION "${UPDATE_ARGUMENTS[@]}")
+"$UPDATE_CMD" "$UPDATE_ACTION" --help | grep -q downloadonly && UPDATE_COMMAND+=(--downloadonly)
 # check for --downloadonly option - if not supported (Debian), fallback to
 # yumdownloader
-if ! "${YUM[@]}" --help | grep -q downloadonly; then
+if ! "$UPDATE_CMD" --help | grep -q downloadonly && ! "$UPDATE_CMD" --version | grep -q dnf5; then
```
marmarek, line 152: While at it, the yumdownload fallback can be dropped - it's relevant for Debian 10, which isn't supported anymore.

package-managers/qubes-download-dom0-updates.sh
```diff
@@ -129,10 +145,11 @@ if [ ${#PKGLIST[@]} -eq 0 ] && [ "$CHECK_ONLY" = "1" ]; then
 fi
 
 # now, we will download something
-YUM_COMMAND=(fakeroot "${YUM[@]}" "$YUM_ACTION" --downloadonly)
+UPDATE_COMMAND=(fakeroot $UPDATE_CMD $UPDATE_ACTION "${UPDATE_ARGUMENTS[@]}")
+"$UPDATE_CMD" "$UPDATE_ACTION" --help | grep -q downloadonly && UPDATE_COMMAND+=(--downloadonly)
```
marmarek, line 149: What if the option is not there? In absence of an alternative block for this, it will try to actually install packages, which will fail. But since yumdownloader fallback is not relevant anymore, I think you can keep adding --downloadonly unconditionally. And if it's missing for some reason, let it fail with a message about it.

marmarek (changes requested):
Besides the few comments, it seems to work

package-managers/qubes-download-dom0-updates.sh
alimirjamali: > I think you can keep adding --downloadonly unconditionally

This will break `qubes-dom0-update --action=search` and `qubes-dom0-update --action=list.`. List and search actions will fail with `--downloadonly` option.

marmarek: Right, so this is a bit smarter detection which action supports the option. Can you add a comment about that (including info that it used to be a global option in DNF4)?

package-managers/qubes-download-dom0-updates.sh
alimirjamali: I deleted the yumdownload block. But will make it a new commit in the same PR. I will also apply shellcheck demands to the same commit.

package-managers/qubes-download-dom0-updates.sh
alimirjamali: Sure. I  will do it

marmarek (approved):

PR #522, opened 2024-09-27
Fix DNF5 repoquery

alimirjamali:
related: https://github.com/QubesOS/qubes-core-admin-client/pull/308

### DIFF
--- a/qubes-rpc/qvm-template-repo-query
+++ b/qubes-rpc/qvm-template-repo-query
@@ -26,11 +26,18 @@ repodir=$(mktemp -d)
 trap 'rm -r "$repodir"' EXIT
 cat > "$repodir/template.repo"
 
+DNF5=false
+if [ "$(readlink /usr/bin/dnf)" = "dnf5" ]; then
+    DNF5=true
+fi
+
 OPTS+=(-y "--setopt=reposdir=${repodir}" --quiet)
 
-# use vendored 'downloadurl' dnf-plugin (fork of 'download' plugin), to print
-# all mirrors
-OPTS+=("--setopt=pluginpath=/usr/lib/qubes/dnf-plugins")
+if ! $DNF5; then
+    # use vendored 'downloadurl' dnf-plugin (fork of 'download' plugin), to print
+    # all mirrors
+    OPTS+=("--setopt=pluginpath=/usr/lib/qubes/dnf-plugins")
+fi
 
 if ! command -v dnf >/dev/null; then
     echo "ERROR: dnf command is missing, please use newer template for your UpdateVM to download templates." >&2
@@ -49,7 +56,11 @@ touch -r "$hashfile" "$repodir/template.repo"
 RET=0
 
 if [ "$1" = "query" ]; then
-    dnf repoquery "${OPTS[@]}" --qf='%{name}|%{epoch}|%{version}|%{release}|%{reponame}|%{downloadsize}|%{buildtime}|%{license}|%{url}|%{summary}|%{description}|' "$SPEC"
+    if $DNF5; then
+        dnf repoquery "${OPTS[@]}" --qf='%{name}|%{epoch}|%{version}|%{release}|%{repoid}|%{downloadsize}|%{buildtime}|%{license}|%{url}|%{summary}|%{description}|\n' "$SPEC"
+    else
+        dnf repoquery "${OPTS[@]}" --qf='%{name}|%{epoch}|%{version}|%{release}|%{repoid}|%{downloadsize}|%{buildtime}|%{license}|%{url}|%{summary}|%{description}|' "$SPEC"
+    fi
     RET="$?"
 elif [ "$1" = "download" ]; then
     # Download/retry algorithm: take mirrors in random order. In this order,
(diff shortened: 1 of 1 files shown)


marmarek:
Downloading templates fail. I guess "downloadurl" plugin is no longer compatible?

alimirjamali:
> Downloading templates fail. I guess "downloadurl" plugin is no longer compatible?

I am not sure. I am experiencing other bugs. With my previous patch (using `%{reponame}` and allowing space in Regex), the `qvm-template-gui` was working fine. It is now broken. I have to investigate.

alimirjamali:
I got that mixed a little.

We have two problems:
> Downloading templates fail. I guess "downloadurl" plugin is no longer compatible?
1. Yes. downloadurl is no longer available. What to do now? Revert to `dnf repoquery package --location` or `dnf download package --url`?

2. The 2nd issue is that dnf5 does not add `\n` to `dnf repoquery --qf=...`.

alimirjamali:
@marmarek 

The patch that I just uploaded should work for both DNF4 and DNF5. But it is ugly and i do not like it. The problem with skipping `downloadurl` is that all downloads will be from one mirror in R4.3 (and Fedora 41 based updatevms). And  I did not manage to find any way to get a list of all urls to shuffle.

alimirjamali:
Created an Issue on upstream:

https://github.com/rpm-software-management/dnf5/issues/1734

alimirjamali:
[This line](https://github.com/rpm-software-management/dnf5/blob/main/dnf5/commands/download/download.cpp#L276) in dnf5 download plugin is what that has to be changed/improved to allow `--all-mirrors`. I am not sure if the upstream DNF5 developers would accept a `--all-mirrors` option. If not, it should be possible to write a fork of dnf5 download plugin to do the same as downloadurl fork of dnf4's download plugin.

In the meantime, the ugly patch I suggested in this PR should allow temporal use of `qvm-template` with Fedora 41 based updatevms.

@marmarek What is your suggestion?

marmarek:
Well, first mirror is better than qvm-template broken completely. But that isn't suitable long term solution, without listing more URLs, fallback to other mirrors won't work (and it is needed sometimes, especially when downloading over Tor).

marmarek:
What about adding a DNF5 "downloadurl" plugin, based on the "download" command upstream? It was pretty simple to make "downloadurl" plugin for DNF4 ;)

marmarek:
In the meantime, I'll merge some of the other PRs and build initial template for testing. It will have broken VM-side of qvm-template still, but will ease testing other stuff (and also allows running this PR on openQA)

alimirjamali:
Update: Path to `download.cpp` and `download.hpp` is done. I am trying to summit the [patch](https://github.com/rpm-software-management/dnf5/pull/1735) to upstream. Maybe DNF5 maintainers would accept it and other people could use it as well.

alimirjamali:
Update: Individual `downloadurl` plugin for DNF5 is ready and tested.

In the next step, we have to decide how long to wait for the upstream maintainers to make a decision if they like the patch or reject it. If they reject it or do not make a decision within the reasonable time, packaging & building instructions for Qubes OS specific plugin is available [here](https://dnf5.readthedocs.io/en/latest/tutorial/plugins/dnf5-plugins.html).

marmarek:
> OpenQA test summary

The qvm-template related tests seems to be green :) 
But there are a few failures related to qubes-vm-update (but that's unrelated to this PR)

alimirjamali:
> > OpenQA test summary
> 
> The qvm-template related tests seems to be green :) (but that's ofc without the --all-mirrors support in DNF5 yet) But there are a few failures related to qubes-vm-update (but that's unrelated to this PR)

That is good. In the meantime, DNF maintainers are in the process of approving the new option. And they asked it to be renamed to `--allmirrors`. And it requires `-q` to filter out repo refreshing info. But overall I guess it would eliminate the necessity for a Qubes specific plugin.

alimirjamali:
> (but that's ofc without the --all-mirrors support in DNF5 yet)

OK. Upstream DNF5 maintainers have accepted [`--allmirrors` PR](https://github.com/rpm-software-management/dnf5/pull/1735) and merged it. It appears that they follow a (bi)weekly new release schedule. I guess it should be readily available very soon.

On the issue of `dnf repoquery --qf=...` not printing `\n` on DNF5, I am not certain if this is a new deliberate feature or a bug. Overall it is not very serious.

marmarek:
> OK. Upstream DNF5 maintainers have accepted [`--allmirrors` PR](https://github.com/rpm-software-management/dnf5/pull/1735) and merged it. It appears that they follow a (bi)weekly new release schedule. I guess it should be readily available very soon.

Can you adjust this PR to work with the updated dnf5 already, including detection if the `--allmirrors` option is there? That would mean 3 cases now:
- dnf4 - custom plugin
- old dnf5 - take just the first url
- new dnf5 - use `--allmirrors` option

alimirjamali:
> Can you adjust this PR to work with the updated dnf5 already,

Of course I can. But I was under the impression that we should wait until DNF5 maintainers make a new release which includes the `--allmirror` patch? The current DNF5 version on Fedora 41 and Fedora Rawhide is 5.2.6.2 from 3 weeks ago which does not include the  `--allmirror` patch.

https://packages.fedoraproject.org/pkgs/dnf5/dnf5/

marmarek:
Given F41 enters final freeze in 4 days, I'm skeptical this update will make it to the F41 release on day 0, more likely it will be an update later...

alimirjamali:
> Given F41 enters final freeze in 4 days, I'm skeptical this update will make it to the F41 release on day 0, more likely it will be an update later...

I rushed in to replying to your previous message after reading the 1st line  (sorry for that. some lack of sleep here). I just adjusted the PR. I have not tested it personally as I am not currently behind my Qubes OS machine. Hopefully the adjustment should work without any issues as it quite simple. I appreciate if you could do a sanity check.

alimirjamali:
This will not work. DNF5 maintainers had asked me for space-separated URLs rather than each on a new line. readarray has to be adjusted as well.

alimirjamali:
@marmarek 
The patch is now ready. I had to readjust it since you made a change last week which resulted to a merge conflict.

qubes-rpc/qvm-template-repo-query
```diff
@@ -49,16 +56,38 @@ touch -r "$hashfile" "$repodir/template.repo"
...
     RET="$?"
 elif [ "$1" = "download" ]; then
     # Download/retry algorithm: take mirrors in random order. In this order,
     # try to download from the first one - if download failed but anything was
     # downloaded - retry from the same one. If download failed and nothing was
     # downloaded, go to the next one. The intention is to retry on interrupted
     # connection, but skip mirrors that are not synchronized yet.
-    urls="$(dnf downloadurl "${OPTS[@]}" --url --all-mirrors "$SPEC" | shuf)"
-    readarray -t urls <<<"$urls"
+    declare -a urls=()
```
marmarek, line 71: Does it make sens if you assign plain string to `urls` below anyway? And then `readarray` will create array later. Only the middle case benefits from this declare, but all cases should be consistent.

marmarek (changes requested):

qubes-rpc/qvm-template-repo-query
alimirjamali: Then I guess the 1st case should use something else for temporary storage of space-separated urls. I will amend it and push the change in a second.

qubes-rpc/qvm-template-repo-query
```diff
@@ -49,16 +56,38 @@ touch -r "$hashfile" "$repodir/template.repo"
...
+    if $DNF5 && dnf download --help | grep -q allmirrors; then
+        # The smartest case. DNF5 on Fedora 41 with --allmirrors patch
+        urls="$(dnf download "${OPTS[@]}" --url --allmirrors "$SPEC")"
+        readarray -d ' ' -t urls <<<"$urls"
+        urls=( $(shuf -e "${urls[@]}") )
+    elif $DNF5; then
+        # The middle case. DNF5 on Fedora 41 before --allmirror patch
+        # TODO: Phase out after DNF5 --allmirrors patch is released
+        url="$(dnf download "${OPTS[@]}" --url "$SPEC")"
+        urls+=($url)
```
marmarek, line 81: Maybe you can simply change this to `urls=($url)`? I don't think there is anything in the `urls` at this point yet?

alimirjamali: OK. Will do it as well

alimirjamali:
@marmarek The two requested changes are applied.

marmarek (approved):

PR #529, opened 2024-10-31
qvm-copy: support `--ignore-symlinks` and honor `--`

DemiMarie:
`--` was stripped from the arguments before being passed to qfile-agent, so `qvm-copy -- --ignore-symlinks` would ignore symlinks instead of copying a file named `--ignore-symlinks`.  To fix this problem, always pass `--` to qfile-agent, preceeded by `--ignore-symlinks` if necessary.  This requires qfile-agent to honor `--` as indicating end of options, so patch it to do that.

### DIFF
--- a/qubes-rpc/qvm-copy
+++ b/qubes-rpc/qvm-copy
@@ -21,7 +21,7 @@
 
 set -e -o pipefail
 
-unset PROGRESS_TYPE OPERATION_TYPE TARGET_TYPE MIN_ARGS FILECOPY_TOTAL_BYTES service scriptdir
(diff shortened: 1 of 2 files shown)


qubes-rpc/qvm-copy
```diff
@@ -66,6 +74,9 @@ export PROGRESS_TYPE=console
 while [ "$#" -gt 0 ]; do
     case $1 in
         (--without-progress) export PROGRESS_TYPE=none; shift;;
+        (--with-progress) export PROGRESS_TYPE=console; shift;;
+        (--ignore-symlinks) ignore_symlinks=; shift;;
```
marmarek, line 78:
```suggestion
        (--ignore-symlinks) ignore_symlinks=true; shift;;
```

or some non-empty other value, make it a bit more readable

marmarek (changes requested):

marmarek (approved):

PR #534, opened 2024-11-10
qubes-fs-tree-check: Detect if a directory is unsafe for display

DemiMarie:
When processing directories, `process_dirent()` detected filenames that are unsafe for display and set `bad` to true, but `bad` was not used to determine the return value of `process_dirent()`.  Instead, the function tail-called `simple_fs_walk()`.  Therefore, directories with unsafe filenames were not detected.  This caused `qubes.Filecopy` to be used instead of `qubes.Filecopy+allow-all-names`.  `qubes.Filecopy` (correctly) rejects the directory.
    
Fix the bug by ensuring that if `bad` is set to true, `process_dirent()` returns true even if the function recurses into `simple_fs_walk()`.  This will cause `qubes.Filecopy+allow-all-names` to be used, which will accept the directory and allow the copy to succeed.
    
Fixes: QubesOS/qubes-issues#9567


DemiMarie:
PipelineRetryFailed

marmarek (approved):

PR #538, opened 2024-12-10
AutostartCondition support

Guiiix:
Support conditional startup of XDG applications by checking the existence of a given file. The check must be configured using the setting ``AutostartCondition=if-exists`` or ``AutostartCondition=unless-exists`` in .desktop files.

### DIFF
--- a/app-menu/qubes-session-autostart
+++ b/app-menu/qubes-session-autostart
@@ -28,9 +28,17 @@ from qubesagent.xdg import launch, find_dropins, \
     load_desktop_entry_with_dropins
 import xdg.BaseDirectory
 import os
+import re
 
 
 QUBES_XDG_CONFIG_DROPINS = '/etc/qubes/autostart'
+
+
+def check_file_existence(path: str) -> bool:
+    if os.path.isabs(path):
+        return os.path.isfile(path)
+    else:
+        return os.path.isfile(os.path.join(xdg.BaseDirectory.xdg_config_home, path))
 
 
 def entry_should_be_started(entry, environments):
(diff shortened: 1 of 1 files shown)


app-menu/qubes-session-autostart
```diff
@@ -28,18 +28,39 @@ from qubesagent.xdg import launch, find_dropins, \
...
 import os
+import re
 
 
 QUBES_XDG_CONFIG_DROPINS = '/etc/qubes/autostart'
 
 
+def check_file_existence(path: str) -> bool:
+    if os.path.isabs(path):
+        return os.path.isfile(path)
```
marmarek, line 39: shouldn't this be `os.path.exists`? or the spec says it needs to be a normal file specifically?

app-menu/qubes-session-autostart
```diff
@@ -28,18 +28,39 @@ from qubesagent.xdg import launch, find_dropins, \
...
+    else:
+        return os.path.isfile(os.path.join(xdg.BaseDirectory.xdg_config_home, path))
```
marmarek, line 41: same here

app-menu/qubes-session-autostart
```diff
@@ -28,18 +28,39 @@ from qubesagent.xdg import launch, find_dropins, \
...
     :type entry: DesktopEntry
     """
     if entry.getHidden():
         return False
+
+    # Handle AutostartCondition keys. Not part of the official XDG specification but documented here:
+    # https://lists.freedesktop.org/archives/xdg/2007-January/007436.html
+    # Only if-exists and unless-exists keywords are supported
+    # Invalid syntax is ignored
+    if entry.get('AutostartCondition', list=True):
```
marmarek, line 56: Is this if necessary? IIUC the loop below would simply iterate 0 times if the entry is not there, no?

marmarek (changes requested):

app-menu/qubes-session-autostart
Guiiix: Right, it's not written that it must necessarily be a file.

app-menu/qubes-session-autostart
Guiiix: Right, I thought it would return None like for dicts, but actually it returns an empty list

marmarek: Yes. But even if it would return None, you could add `|| []` and all would be fine.

marmarek:
Can you squash fixes into the original commit?

Guiiix:
Yes!

marmarek (approved):

app-menu/qubes-session-autostart
Guiiix: Thank you for the tip

PR #541, opened 2024-12-23
Relabel / and /rw if needed

DemiMarie:
Creating /.autorelabel must cause a Qubes OS VM to relabel everything, as otherwise users will not be able to troubleshoot their systems and upstream packages that create it will break.  However, it was ignored, so fix that.

Furthermore, relabel the filesystem of a TemplateBasedVM whenever its TemplateVM has been relabeled since the TemplateBasedVM was.  This ensures that policy changes propagate to TemplateBasedVMs too.


marmarek:
This fails to build...

But also, shouldn't this reference #9663 ?
And additionally, I think we need the updater to support handling reboot for relabeling after an update that requires it...

DemiMarie:
I meant for this to be a draft because I hadn’t tested it, hence the build failure.

DemiMarie:
I didn’t reference #9663 because I don’t know if this actually fixes that issue.

init/relabel-rw.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash --
```
marmarek, line 1: What bash features you use here? I'd prefer /bin/sh (and consequently shellcheck noticing non-POSIX features) in scritpts that don't absolutely need bash (for example to have proper arrays).

init/relabel-rw.sh
```diff
@@ -0,0 +1,6 @@
...
+set -eu
+if [ /.qubes-relabeled -nt /rw/.autorelabel ]; then
```
marmarek, line 3: Theoretically this check is not enough, in case you switch between templates (for example switch from F41 back to F40). For that it would need to compare some hash on the policy or similar, not just the freshness. But in practice, I don't expect issues (most labels that change on updates are outside of /rw).

marmarek: For now, can you change it to file_contexts timestamp check? And later we probably will need a proper update of only changed labels...

marmarek:
shellcheck needs a comment near sourcing /etc/selinux/config

marmarek (approved):

PR #542, opened 2024-12-23
Pull in qubes-core-agent-network-manager when installing NetworkManager

DemiMarie:
The documentation says to install qubes-core-agent-network-manager, but the system can just do it for the user by pulling it in via an RPM rich dependency.

Fixes: QubesOS/qubes-issues#9664


marmarek:
PipelineRetry

marmarek (approved):

PR #545, opened 2024-12-24
Add missing .gitignore entries

DemiMarie:
No functional change intended.  This will help developers realize when there is a file in their working tree that should not be there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,14 @@ deb/*
 __pycache__
 debian/changelog.*
 pkgs
-misc/mime/[^p]*
-misc/mime/mimeapps.list
-misc/mime/mimeinfo.cache
+/misc/mime/[^p]*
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -9,6 +9,15 @@ deb/*
...
-misc/mime/mimeapps.list
-misc/mime/mimeinfo.cache
+/misc/mime/[^p]*
+/misc/applications/mimeapps.list
+/misc/applications/mimeinfo.cache
+/misc/applications/defaults.list
+/qubes-rpc/bin-qfile-unpacker
+/qubes-rpc/qubes-fs-tree-check
+# Created by builder
+/hash
```
marmarek, line 19: where?

DemiMarie: I definitely had the `hash` file in a bunch of repositories.  Maybe an old builder created it?

PR #547, opened 2025-01-14
Remove `hostname` dependency for Archlinux

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/9710

### DIFF
--- a/archlinux/PKGBUILD.install
+++ b/archlinux/PKGBUILD.install
@@ -114,7 +114,7 @@ update_qubesconfig() {
         if ! grep -q localhost /etc/hosts; then
 
           cat <<EOF > /etc/hosts
-127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4 $(hostname)
+127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4 $(hostnamectl hostname)
 ::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
 EOF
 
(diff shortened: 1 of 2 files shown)


marmarek:
There is more: hostname is also set on the VM startup: https://github.com/QubesOS/qubes-core-agent-linux/blob/main/vm-systemd/qubes-early-vm-config.sh#L22

alimirjamali:
> There is more: hostname is also set on the VM startup: https://github.com/QubesOS/qubes-core-agent-linux/blob/main/vm-systemd/qubes-early-vm-config.sh#L22

This one uses the `SET NAME` feature of hostname command. I will modify the bash script to check for the existence of `/usr/bin/hostname` and revert to `hostnamectl` if it does not exist. I will also change `cat /etc/hostname` to `hostnamectl hostname` in the original PR as suggested by @justasic

marmarek:
Now setting hostname is broken on Fedora (at least): https://openqa.qubes-os.org/tests/126217#step/screenlocker_idle_watch/10 (see that hostname remained "fedora", should be "work).

marmarek:
Logs say:
```
[2025-01-21 05:39:58] [    4.550584] hostnamectl[512]: [0;1;31mFailed to connect to system scope bus via local transport: No such file or directory[0m
```
I guess it's too early for hostnamed service...

alimirjamali:
> I guess it's too early for hostnamed service...

Maybe `$(echo "$name" > /etc/hostname)` could work but it is a little bit ugly.

marmarek:
But will that be enough? Something needs to actually make the `sethostname` syscall...
Maybe writing to `/proc/sys/kernel/hostname` will be enough?

alimirjamali:
> But will that be enough? Something needs to actually make the `sethostname` syscall... Maybe writing to `/proc/sys/kernel/hostname` will be enough?

Does not work with root. I get Permission denied on `echo something > /proc/sys/kernel/hostname`.

alimirjamali:
I modified it a little bit (only try alternative solution if `hostname binary does not exist). And trying to write `$name` to `/etc/hostname`, hoping systemd would pick it up on archlinux

marmarek:
> Does not work with root. I get Permission denied on `echo something > /proc/sys/kernel/hostname`.

You sure you haven't done the classic `sudo echo` pitfall? Writing there does work for me...

marmarek:
Anyway, I guess /etc/hostname needs to be written to too

alimirjamali:
> Anyway, I guess /etc/hostname needs to be written to too (if exists)

I amended the PR to write to both. Hope it works well this time.

marmarek (approved):

PR #548, opened 2025-01-16
Fix qvm-copy behavior with symlinks

zaz:
Previously, `qvm-copy` had a couple of issues with symlinks:

1. A trailing slash on the source could cause errors:

````
$ qvm-copy-to-vm untrusted dir/
qubes-fs-tree-check: Refusing to copy unsafe symbolic link "dir//my-symlink"
````

but `qvm-copy-to-vm untrusted dir` worked fine.

2. `qvm-copy --ignore-symlinks` could still throw `qubes-fs-tree-check: Refusing to copy unsafe symbolic link` . If we are ignoring symlinks, `qubes-fs-tree-check` should not check for them.

## Notes

Running `qubes-fs-tree-check --ignore-symlinks --allow-symlinks` does not seem to cause an issue.

### DIFF
--- a/qubes-rpc/qvm-copy
+++ b/qubes-rpc/qvm-copy
@@ -97,6 +97,7 @@ fi
 SOURCE="${@%/}"
 
 if FILECOPY_TOTAL_BYTES=$("$scriptdir/qubes/qubes-fs-tree-check" \
+    ${ignore_symlinks+--ignore-symlinks} \
     --allow-symlinks --allow-directories --machine -- "${SOURCE}"); then
     service=qubes.Filecopy
 else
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: qubes-rpc/qvm-copy [hunk 1]
```diff
@@ -93,8 +93,12 @@ else
   VM="@default"
 fi
 
+# strip trailing slash from source directory
+SOURCE="${@%/}"
```
marmarek, line 97: This breaks handling multiple paths. But you can keep it being an array with `SOURCE=( "${@%/}" )`, and later use "${SOURCE[@]}".

marmarek (changes requested):

zaz:
@marmarek Thanks for catching that! How does it look now?

marmarek (approved):

PR #551, opened 2025-01-28
New custom-persist feature

Guiiix:
This PR adds a new feature ``custom-persist`` described in https://github.com/QubesOS/qubes-issues/issues/1006

### DIFF
--- a/init/functions
+++ b/init/functions
@@ -57,6 +57,10 @@ is_appvm() {
     [ "$(qubes_vm_type)" = "AppVM" ]
 }
 
+is_custom_persist_enabled() {
+    [ -f "/var/run/qubes-service/custom-persist" ]
+}
+
 is_proxyvm() {
     [ "$(qubes_vm_type)" = "ProxyVM" ]
 }
--- a/init/functions
+++ b/init/functions
@@ -23,7 +23,9 @@ qsvc() {
 }
 
 under_systemd() {
-    pidof systemd >/dev/null 2>&1
+    local init_command_name
+    read -r init_command_name < /proc/1/comm
+    [ "$init_command_name" = "systemd" ]
 }
 
 systemd_version_changed() {
--- a/init/functions
+++ b/init/functions
@@ -255,12 +255,10 @@ initialize_home() {
 
 disable_persistent_home() {
     echo "Disabling persistent /home"
-    mkdir /run/systemd/system/home.mount.d
-    echo -e '[Mount]\nWhat=/home' > /run/systemd/system/home.mount.d/30_qubes.conf
+    touch /var/run/qubes/disable_persistent_home_dir
 }
 
 disable_persistent_usrlocal() {
     echo "Disabling persistent /usr/local"
-    mkdir /run/systemd/system/usr-local.mount.d
-    echo -e '[Mount]\nWhat=/usr/local' > /run/systemd/system/usr-local.mount.d/30_qubes.conf
+    touch /var/run/qubes/disable_persistent_usrlocal_dir
 }
--- a/init/functions
+++ b/init/functions
@@ -250,3 +250,15 @@ initialize_home() {
         for waitpid in $waitpids ; do wait "$waitpid" ; done ; waitpids=
     done
 }
+
+disable_persistent_home() {
+    echo "Disabling persistent /home"
+    mkdir /run/systemd/system/home.mount.d
+    echo -e '[Mount]\nWhat=/home' > /run/systemd/system/home.mount.d/30_qubes.conf
+}
+
+disable_persistent_usrlocal() {
+    echo "Disabling persistent /usr/local"
+    mkdir /run/systemd/system/usr-local.mount.d
+    echo -e '[Mount]\nWhat=/usr/local' > /run/systemd/system/usr-local.mount.d/30_qubes.conf
+}
--- a/qubesagent/firewall.py
+++ b/qubesagent/firewall.py
@@ -86,10 +86,15 @@ class FirewallWorker(object):
             return []
         return ips.decode().split()
 
+    def is_custom_persist_enabled(self) -> bool:
+        """Check if the feature custom-persist is enabled on the current VM"""
+        return os.path.isfile('/var/run/qubes-service/custom-persist')
+
     def run_firewall_dir(self):
         """Run scripts dir contents, before user script"""
-        script_dir_paths = ['/etc/qubes/qubes-firewall.d',
-                            '/rw/config/qubes-firewall.d']
+        script_dir_paths = ['/etc/qubes/qubes-firewall.d']
+        if not self.is_custom_persist_enabled():
+            script_dir_paths.append('/rw/config/qubes-firewall.d')
         for script_dir_path in script_dir_paths:
             if not os.path.isdir(script_dir_path):
                 continue
@@ -328,7 +333,8 @@ class FirewallWorker(object):
         self.terminate_requested = False
         self.init()
         self.run_firewall_dir()
-        self.run_user_script()
+        if not self.is_custom_persist_enabled():
+            self.run_user_script()
         self.sd_notify('READY=1')
         self.qdb.watch('/qubes-firewall/')
         self.qdb.watch('/connected-ips')
--- a/vm-systemd/bind-dirs.sh
+++ b/vm-systemd/bind-dirs.sh
@@ -144,6 +144,39 @@ if is_custom_persist_enabled; then
   while read -r qubes_persist_entry; do
     [[ "$qubes_persist_entry" =~ =\ (.*)$ ]] || continue
     target="${BASH_REMATCH[1]}"
+
+    # if the first char is not a slash, options should be extracted from
+    # the value
+    if [[ "$target" != /* ]]; then
+      resource_type="$(echo "$target" | cut -d':' -f1)"
+      owner="$(echo "$target" | cut -d':' -f2)"
+      group="$(echo "$target" | cut -d':' -f3)"
+      mode="$(echo "$target" | cut -d':' -f4)"
+      path="$(echo "$target" | cut -d':' -f5-)"
+
+      if [ -z "$path" ] || [[ "$path" != /* ]]; then
+        echo "Skipping invalid custom-persist value '${target}'" >&2
+        continue
+      fi
+
+      # create resource if it does not exist
+      if ! [ -e "${path}" ] && ! [ -e "/rw/bind-dirs${path}" ]; then
+        if [ "$resource_type" = "file" ]; then
+          # for files, we need to create parent directories
+          parent_directory="$(dirname "$path")"
+          [ -d "$parent_directory" ] || mkdir -p "${parent_directory}"
+          touch "${path}"
+        elif [ "$resource_type" = "dir" ]; then
+          mkdir -p "${path}"
+        else
+          echo "Invalid entry ${target}, skipping"
+          continue
+        fi
+        chown "$owner":"$group" "${path}"
+        chmod "$mode" "${path}"
+      fi
+      target="$path"
+    fi
     [[ "$target" =~ ^(\/home|\/usr\/local)$ ]] && continue
     binds+=( "$target" )
   done <<< "$(qubesdb-multiread /persist/)"
--- a/vm-systemd/bind-dirs.sh
+++ b/vm-systemd/bind-dirs.sh
@@ -159,21 +159,24 @@ if is_custom_persist_enabled; then
         continue
       fi
 
+      rw_path="/rw/bind-dirs${path}"
       # create resource if it does not exist
-      if ! [ -e "${path}" ] && ! [ -e "/rw/bind-dirs${path}" ]; then
+      if ! [ -e "${path}" ] && ! [ -e "$rw_path" ]; then
         if [ "$resource_type" = "file" ]; then
           # for files, we need to create parent directories
-          parent_directory="$(dirname "$path")"
+          parent_directory="$(dirname "$rw_path")"
+          echo "custom-persist: pre-creating file ${rw_path} with rights ${owner}:${group} ${mode}"
           [ -d "$parent_directory" ] || mkdir -p "${parent_directory}"
-          touch "${path}"
+          touch "${rw_path}"
         elif [ "$resource_type" = "dir" ]; then
-          mkdir -p "${path}"
+          echo "custom-persist: pre-creating directory ${rw_path} with rights ${owner}:${group} ${mode}"
+          mkdir -p "${rw_path}"
         else
           echo "Invalid entry ${target}, skipping"
           continue
         fi
-        chown "$owner":"$group" "${path}"
-        chmod "$mode" "${path}"
+        chown "$owner":"$group" "${rw_path}"
+        chmod "$mode" "${rw_path}"
       fi
       target="$path"
     fi
--- a/vm-systemd/bind-dirs.sh
+++ b/vm-systemd/bind-dirs.sh
@@ -129,7 +129,7 @@ bind_dirs() {
 
       # Bind the fso.
       echo "Bind mounting $fso_rw onto $fso_ro" >&2
-      mount --bind "$fso_rw" "$fso_ro"
+      mount --bind -o x-gvfs-hide "$fso_rw" "$fso_ro"
    done
 }
 
--- a/vm-systemd/bind-dirs.sh
+++ b/vm-systemd/bind-dirs.sh
@@ -133,6 +133,19 @@ bind_dirs() {
    done
 }
 
+mk_parent_dirs() {
+  local target="$1"
+  local owner="$2"
+  local group="$3"
+  local depth="$4"
+  [[ "$depth" -gt 100 ]] && echo "Maximum recursion depth reached" >&2 && return 1
+  [ -e "$target" ] && return 0
+  mk_parent_dirs "$(dirname "$target")" "$owner" "$group" "$(( depth + 1 ))" || return 1
+  mkdir "$target" || return 1
+  chown "$owner":"$group" "$target" || return 1
+  return 0
+}
+
 main() {
    prerequisite "$@"
    init "$@"
@@ -185,11 +198,19 @@ if is_custom_persist_enabled; then
           # for files, we need to create parent directories
           parent_directory="$(dirname "$rw_path")"
           echo "custom-persist: pre-creating file ${rw_path} with rights ${owner}:${group} ${mode}"
-          [ -d "$parent_directory" ] || mkdir -p "${parent_directory}"
+          if [ ! -d "$parent_directory" ]; then
+            if ! mk_parent_dirs "$parent_directory" "$owner" "$group"; then
+              echo "Unable to create ${rw_path} parent dirs, skipping"
+              continue
+            fi
+          fi
           touch "${rw_path}"
         elif [ "$resource_type" = "dir" ]; then
           echo "custom-persist: pre-creating directory ${rw_path} with rights ${owner}:${group} ${mode}"
-          mkdir -p "${rw_path}"
+          if ! mk_parent_dirs "$rw_path" "$owner" "$group"; then
+            echo "Unable to create ${rw_path} parent dirs, skipping"
+            continue
+          fi
         else
           echo "Invalid entry ${target}, skipping"
           continue
--- a/vm-systemd/bind-dirs.sh
+++ b/vm-systemd/bind-dirs.sh
@@ -118,7 +118,12 @@ main() {
 }
 
 binds=()
-for source_folder in /usr/lib/qubes-bind-dirs.d /etc/qubes-bind-dirs.d /rw/config/qubes-bind-dirs.d ; do
+sources=( "/usr/lib/qubes-bind-dirs.d" "/etc/qubes-bind-dirs.d" )
+if [ ! -f "/var/run/qubes-service/custom-persist" ]; then
+    sources+=( "/rw/config/qubes-bind-dirs.d" )
+fi
+
+for source_folder in "${sources[@]}"; do
    true "source_folder: $source_folder"
    if [ ! -d "$source_folder" ]; then
       continue
--- a/vm-systemd/bind-dirs.sh
+++ b/vm-systemd/bind-dirs.sh
@@ -91,7 +91,11 @@ bind_dirs() {
          if [ ! -e "$fso_ro" ]; then
             ## Create empty file or directory if path exists in /rw to allow to bind mount none existing files/dirs.
             test -d "$fso_rw" && mkdir --parents "$fso_ro"
-            test -f "$fso_rw" && touch "$fso_ro"
+            if [ -f "$fso_rw" ]; then
+              parent_directory="$(dirname "$fso_ro")"
+              test -d "$parent_directory" || mkdir --parents "$parent_directory"
+              touch "$fso_ro"
+            fi
          fi
       else
          if [ -d "$fso_ro" ] || [ -f "$fso_ro" ]; then
--- a/vm-systemd/bind-dirs.sh
+++ b/vm-systemd/bind-dirs.sh
@@ -107,10 +107,12 @@ bind_dirs() {
       if [ -d "$fso_rw" ] || [ -f "$fso_rw" ]; then
          if [ ! -e "$fso_ro" ]; then
             ## Create empty file or directory if path exists in /rw to allow to bind mount none existing files/dirs.
-            test -d "$fso_rw" && mkdir --parents "$fso_ro"
+            # shellcheck disable=SC2046
+            test -d "$fso_rw" && mk_parent_dirs "$fso_ro" $(stat --printf "%U %G" "$fso_rw")
             if [ -f "$fso_rw" ]; then
               parent_directory="$(dirname "$fso_ro")"
-              test -d "$parent_directory" || mkdir --parents "$parent_directory"
+              # shellcheck disable=SC2046
+              test -d "$parent_directory" || mk_parent_dirs "$parent_directory" $(stat --printf "%U %G" "$fso_rw")
               touch "$fso_ro"
             fi
          fi
--- a/vm-systemd/bind-dirs.sh
+++ b/vm-systemd/bind-dirs.sh
@@ -135,6 +135,16 @@ for source_folder in "${sources[@]}"; do
    done
 done
 
+# read binds in QubesDB if custom-persist feature is enabled
+if is_custom_persist_enabled; then
+  while read -r qubes_persist_entry; do
+    [[ "$qubes_persist_entry" =~ =\ (.*)$ ]] || continue
+    target="${BASH_REMATCH[1]}"
+    [[ "$target" =~ ^(\/home|\/usr\/local)$ ]] && continue
+    binds+=( "$target" )
+  done <<< "$(qubesdb-multiread /persist/)"
+fi
+
 main "$@"
 
 true "OK: END."
--- a/vm-systemd/bind-dirs.sh
+++ b/vm-systemd/bind-dirs.sh
@@ -29,6 +29,8 @@ shopt -s nullglob dotglob
 # shellcheck source=init/functions
 source /usr/lib/qubes/init/functions
 
+readonly DEFAULT_RW_BIND_DIR="/rw/bind-dirs"
+
 prerequisite() {
    if is_fully_persistent ; then
       echo "No TemplateBasedVM/DisposableVM detected. Exiting."
@@ -37,7 +39,7 @@ prerequisite() {
 }
 
 init() {
-   [ -n "$rw_dest_dir" ] || rw_dest_dir="/rw/bind-dirs"
+   [ -n "$rw_dest_dir" ] || rw_dest_dir="$DEFAULT_RW_BIND_DIR"
    [ -n "$symlink_level_max" ] || symlink_level_max="10"
    mkdir --parents "$rw_dest_dir"
 }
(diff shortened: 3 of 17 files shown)


init/functions
```diff
@@ -23,7 +23,7 @@ qsvc() {
 }
 
 under_systemd() {
-    pidof systemd >/dev/null 2>&1
+    [ "$(cat /proc/1/comm)" = "systemd" ]
```
DemiMarie, line 26: Please user the shell builtin `read` rather than `cat`.

init/functions
```diff
@@ -246,3 +250,21 @@ initialize_home() {
...
     done
 }
+
+disable_persistent_home() {
+  echo "Disabling persistent /home"
+  mkdir /run/systemd/system/home.mount.d
+  cat << EOF > /run/systemd/system/home.mount.d/30_qubes.conf
+[Mount]
+What=/home
+EOF
```
DemiMarie, line 260: Just use `printf` or `echo`.

init/functions
```diff
@@ -246,3 +250,21 @@ initialize_home() {
...
+}
+
+disable_persistent_usrlocal() {
+  echo "Disabling persistent /usr/local"
+  mkdir /run/systemd/system/usr-local.mount.d
+  cat << EOF > /run/systemd/system/usr-local.mount.d/30_qubes.conf
+[Mount]
+What=/usr/local
+EOF
```
DemiMarie, line 269: Same here.

qubesagent/firewall.py
```diff
@@ -86,10 +86,15 @@ def get_connected_ips(self, family):
             return []
         return ips.decode().split()
 
+    def is_custom_persist_enabled(self) -> bool:
+        """Check if the feature custom-persist is enabled on the current VM"""
+        return os.path.isfile('/var/run/qubes-service/custom-persist')
```
DemiMarie, line 91: I suggest using an explicit call to `stat` and catching only `FileNotFoundError`.

vm-systemd/mount-dirs.sh
```diff
@@ -10,9 +10,33 @@ set -e
...
+if is_custom_persist_enabled; then
+  mount_home=false
+  mount_usr_local=false
+
+  while read -r qubes_persist_entry; do
+    [ -z "${qubes_persist_entry}" ] && continue
+    mount_point="$(qubesdb-read /persist/"${qubes_persist_entry}")"
+    [ "$mount_point" = "/home" ] && mount_home=true
+    [ "$mount_point" = "/usr/local" ] && mount_usr_local=true
+  done <<< "$(qubesdb-list /persist/)"
```
DemiMarie, line 22: I would use `qubesdb-multiread` here, or (preferred) write this in Python rather than in shell (unless it would slow startup too much).

Guiiix: It would require rewriting all the mount-dirs script (and init/functions) in Python no?

qubesagent/firewall.py
Guiiix: Okay, may I ask why is it better (or safer? faster?) than ``os.path.exists`` for example?

marmarek: It isn't really. I think Demi doesn't believe others do proper error handling ;) (there may be corner cases where os.path.isfile etc will return False on I/O error or such, but that isn't really relevant here)

DemiMarie: You think correctly @marmarek.

marmarek:
> `/home/user/some-dir` not present

Yes, I think bind-dirs does not create dirs on its own, they need to exist beforehand. This is especially relevant since by looking at just the path, you don't have enough info to create it - it can be either a file or a directory, you don't know what permissions to use, owner etc...

But maybe this can be extended with some defaults? For example if doesn't exist, then create as a dir, with owner same as the parent directory, and default mode of 755? Or extend the syntax to allow specify those details...

marmarek:
> `fedora-41-xfce` wouldn't open GUI applications any longer - and yes, I did add `custom-persist` just to check if this would start working and it did, but with some side-effects: the app qube would have the template's home instead. I even though it had actually the template's private volume mounted.

Can you check why? The default /home from the template should be good enough for normal applications to start...

marmarek:
The selinux messages are red herring, it's related to GUI session failing to start, so application is started outside of it. The actual problem is somewhere about GUI agent - see either qubes-gui-agent logs in journal or /home/user/.xsession-errors.

Guiiix:
Hi @deeplow! Thank you for your message. I'm very happy to be part to this adventure with you all :)

I tried to reproduce on release r4.2 (commits backported to branch release 4.2) but I didn't get any error on the template:
  * ``/home/user/somedir`` not mounted: as Marek said, this is the expected behavior regarding the current bind_dirs script. You would have the same result without custom-persist if you had added this directory to ``/rw/config/qubes-bind-dirs.d`` but it could be relevant to do some changes, at least for this feature.
  * fedora-40-xfce template's GUI is working well


Did both``qubes-bind-dirs.service`` and ``qubes-mount-dirs.service`` started well? ``systemctl --failed`` shows anything abnormal?

marmarek:
FWIW I also didn't reproduced the GUI apps failure, but I tried it in AppVM, not the template itself.

I added integration tests here: https://github.com/QubesOS/qubes-core-admin/pull/654/files#diff-19c2f2ac398e72a651b363ef17c05d7f0348b8bed0cbbf5c938265e09c76e463R57-R118

There are a couple of things going on. First, it does fail (as expected) on directories that don't exist initially (one may expect /home/user/Downloads to exists already, but in practice it's created on first boot - and that happens after bind-dirs call). But there is another thing - `/usr/local/bin/true-copy` did persisted but it shouldn't. I see the whole /usr/local being still bind-mounted with /rw/usrlocal...
FWIW that's on debian-12-xfce template.
In the log I see first `systemd[1]: Mounting usr-local.mount - /usr/local...` and only later `Disabling persistent /usr/local`.
And also, the same happens for /home, but here I see `systemd[1]: mounting  home.mount - /home...` after `Disabling persistent /home` message. But here it seems it mounted rootfs (/dev/xvda3, not /dev/xvdb) which I guess is okay.
So, I think at the very least some ordering is wrong, but maybe also missing `systemctl daemon-reload` call?

Guiiix:
I tried many things to reproduce but I was not able to. Maybe I'm doing wrong. I did the following (I'm running a 4.2-stable version of Qubes):

- new template ``debian-persist-tpl`` based on ``debian-12-xfce``
- started the new template, added 4.3 bookworm-testing in apt sources et removed 4.2
- ran ``apt dist-upgrade``
- built linux-core-agent.deb using qubes builder v2 and repo https://github.com/Guiiix/qubes-core-agent-linux, branch ``custom-persist`` (tried also with https://gitlab.com/QubesOS/qubes-core-agent-linux/-/commits/pr-551)
- installed ``qubes-core-agent`` package on ``debian-persist-tpl``
- Created a new ``debian-persist`` APP VM based on ``debian-persist-tpl``. ``qvm-service --enable debian-persist custom-persist``
- Started ``debian-persist``

Rootfs is not mounted for ``/home`` nor ``/usr/local``:
```
root@debian-persist:/home/user# mount | grep -E '(home|local)'
/dev/xvda3 on /home type ext4 (rw,nosuid,nodev,noatime,discard)
/dev/xvda3 on /usr/local type ext4 (rw,noatime,discard)
```

Order seems to be good in logs:
```
Feb 16 12:13:17 localhost mount-dirs.sh[362]: Disabling persistent /usr/local
...
Feb 16 12:13:17 localhost systemd[1]: systemd-repart.service - Repartition Root Disk was skipped because no trigger condition checks were met.
Feb 16 12:13:18 localhost systemd[1]: home.mount: Directory /home to mount over is not empty, mounting anyway.
Feb 16 12:13:18 localhost systemd[1]: Mounting home.mount - /home...
Feb 16 12:13:18 localhost systemd[1]: Mounting usr-local.mount - /usr/local...
Feb 16 12:13:18 localhost systemd[1]: Mounted home.mount - /home.
Feb 16 12:13:18 localhost systemd[1]: Mounted usr-local.mount - /usr/local.
Feb 16 12:13:18 localhost systemd[1]: Reached target local-fs.target - Local File Systems.
Feb 16 12:13:18 localhost systemd[1]: Starting apparmor.service - Load AppArmor profiles...
Feb 16 12:13:18 localhost systemd[1]: Starting qubes-bind-dirs.service - Mount configured bind file
```

marmarek:
Could be non-deterministic. What happens if you set vcpus to 1 for this qube?

marmarek:
Could be non-deterministic. What happens if you set vcpus to 1 for this qube?

marmarek:
Or maybe it's more likely to fail if you add more custom-persist dirs?

marmarek:
In any case, I don't see anything that would guarantee /usr/local and /home being mounted only after bind-dirs is called. I think explicit `After=` is needed (in the `[Unit]` section) there.

Guiiix:
You're right, it's probably non-deterministic. Even with 1vcpu and a lot of custom persist directories, it's working well on my VM but something is missing in the  ``Before=`` section of ``qubes-mount-dirs.service``.

If we want to keep the same behavior as before when this feature is disabled, we don't want ``/home`` and ``/usr/local`` to be mounted after ``qubes-bind-dirs`` as this is ``mount-dirs.sh`` that was mouting them just before executing ``bind-dirs.sh``. Instead, we want:

1. ``qubes-mount-dirs.service``: will create droppins to disable or not ``/home`` and ``/usr/local`` persistence and do a daemon-reload
2. mount ``/home`` and ``/usr/local``
3. Run ``qubes-bind-dirs.service``

``qubes-bind-dirs.service`` has a ``Requires=`` and ``After=`` dependencies with ``qubes-bind-dirs``, ``home.mount`` and ``usr-local.mount`` so if think we are good but ``qubes-mount-dirs.service`` has only a ``Before=`` with ``home.mount``, nothing with ``usr-local.mount``.

I guess this is why you see ``Disabling /usr/local`` after ``systemd[1]: Mounting usr-local.mount - /usr/local...``.

marmarek:
> given that the template does not have `custom-persist`

Template without `custom-persist` will not touch `/home` on the root volume, it will mount template's private /home over it. 

But in any case, /home should be initialized here: https://github.com/QubesOS/qubes-core-agent-linux/pull/551/files#diff-85250c31cd962d4a997ba66dbe7d9df0687c1294ce96bb939bb1f3935ab77b41R31

Do you have logs from bind-dirs service start? Maybe it failed some some reason?

marmarek:
> Yes, I think bind-dirs does not create dirs on its own, they need to exist beforehand. This is especially relevant since by looking at just the path, you don't have enough info to create it - it can be either a file or a directory, you don't know what permissions to use, owner etc...
> 
> But maybe this can be extended with some defaults? For example if doesn't exist, then create as a dir, with owner same as the parent directory, and default mode of 755? Or extend the syntax to allow specify those details...

Looking at how this feature can be used, I think we need this part too. The more I think about it, the more I lean towards explicit metadata option, specifically:
Have a syntax for specifying all that; that syntax could be for example a path prefix like `dir:user:user:0750:` (directory, owner `user:user`, mode 0750), or `file:root:root:0600:`. Using a prefix means it can be distinguished from normal path that should start with `/`, so there is no issue with interpreting metadata as a path. For example `dir:user:user:0750:/home/user/Downloads`.
In case the file/directory already exists, I'm not sure what should happen. I'm leaning towards keeping its mode/owner unchanged. In other words - info from the prefix is used only if the path doesn't exist yet.

marmarek:
Handling runtime reload might be a nice addition, but I'd prefer to keep it for future extension, not this PR. There are several corner cases there, which is a reason why it isn't supported for normal bind-dirs (by restarting the service manually) either.

Guiiix:
Couldn't we put all of that on more global feature like "advanced persistence"?
An APP VM might be configured to use either "basic persistence" or "advanced persistence". When advanced persistence would be activated, instead of disabling everything, we could let the user choose what he wants to keep or not (for example, he may want to keep the default mount points or not, keep the execution of /rw/config/rc.local or not, keep the /rw bind dirs configuration etc).
This would give the user a total control of the AppVM behavior.

marmarek:
> may want to keep the default mount points or not

This you already implemented, by simply adding `/home` and/or `/usr/local` to custom persist.

marmarek:
And when using custom-persist feature, the bind dirs config in /rw is kinda redundant anyway.

Guiiix:
With this commit, files and directories are automatically created if they don't exist in the template and /rw. If no metadata is filled, the default behavior is kept. 

I also added the creation of parent directories when a file has to be mounted as the service was crashing when trying to touch the file : 

https://github.com/QubesOS/qubes-core-agent-linux/pull/551/commits/e0003fc1bfb3ce435df3de02cd02175598980378#diff-2e803fe023ac5278e31a6428053ccfedcd458e76b2fa246969efa49429242351L94

This applies even if custom-persist is not enabled.

Guiiix:
Yes, I saw your previous message, I'll take a look today. :)

marmarek:
Something is wrong with the /usr/local handling. First I see:
```
[2025-02-20 15:54:07] [    3.539350] mount-dirs.sh[392]: Virgin boot of the VM: populating /rw/usrlocal from /usr/local.orig
```
and `/usr/local.orig` should include standard `bin`, `lib` etc directories.
But then:
```
[2025-02-20 15:54:07] [    3.896212] bind-dirs.sh[445]: /usr/local/lib is neither a directory nor a file and the path does not exist below /rw, skipping.
```
Maybe it's looking only in `/rw/bind-dirs`? IMO there should be a special case for paths under `/usr/local` and `/home` look for persistent version in `/rw/usrlocal` and `/rw/home` respectively? IMO that would be logical since adding `/home` or `/usr/local` to custom-persist list enables mounting those standard locations. And also it would be more logical when enabling/disabling the feature, as the files/directories selected to be persistent would remain the same, as stored in `/rw/usrlocal` / `/rw/home` (instead of swapping to their alternative versions in `/rw/bind-dirs`).

Guiiix:
I try to reproduce the issue with Thunar but I wasn't able to... Maybe I'm doing wrong? I did the following:

- clone the current template ``fedora-41-xfce`` to ``custom-persist-tpl``
- edit yum repos in the new template to enable r4.3-current and r4.3-current-testing
- update
- update ``qubes-core-agent`` and ``qubes-core-agent-systemd`` with this version.

I added some custom-persist directories. ``mount`` tells me they are bind to ``xvdb``:
```
When masking the service:
  * socket fails to start before the service couldn't be loaded
 
When masking the socket or both
  * no service failed nor error in journal
```

But I don't see anything in Thunar. Am I missing something?

---

For the behavior of ``bind-dirs.sh`` yes this is exactly what's happening and I agree, it would be logical to check under ``/rw/userlocal`` and ``/rw/home``. We can also create 2 symlinks in ``/rw/bind-dirs`` but maybe it's a bit ugly...

Should this check been done even if custom persist is not activated? I know this would be a no sense to add a bind dir in ``/home`` or ``/usr/local``  without the feature enabled, but if that were done, the same error would be raised.

marmarek:
> Should this check been done even if custom persist is not activated? I know this would be a no sense to add a bind dir in `/home` or `/usr/local` without the feature enabled, but if that were done, the same error would be raised.

Yes, I think it makes sense to be consistent in that case (without custom-persist, such entries would be more or less no-op, besides extra mount line).

Guiiix:
Fixed in last 2 commits.

There is still a limitation (or even a regression) if one declares a bind dir/file in ``/home`` or ``/usr/local`` and if it exists in the template. In that particular case, ``bind_dirs()`` will copy files  to ``/rw/bind-dirs`` instead of ``/rw/home`` or ``/rw/usrlocal`` : 

https://github.com/QubesOS/qubes-core-agent-linux/blob/ff6742c2b5632d48b98cba3da8e1106d9292c2a1/vm-systemd/bind-dirs.sh#L121

This would cause the service to fail because the mount command will try to mount a non-existing path: https://github.com/QubesOS/qubes-core-agent-linux/blob/ff6742c2b5632d48b98cba3da8e1106d9292c2a1/vm-systemd/bind-dirs.sh#L130

marmarek:
It fails on Whonix, but works on plain fedora/debian, not sure what is the difference, the Downloads directory should be the same...
```
[2025-02-23 09:16:21] [    4.516303] bind-dirs.sh[458]: /var/spool/cron is not a symlink
[2025-02-23 09:16:21] [    4.527679] enable-firewall[482]: 2025-02-23 14:16:21 - whonix-workstation-firewall - OK: Skipping firewall mode detection since already set to 'full'.
[2025-02-23 09:16:21] [    4.528449] bind-dirs.sh[458]: Initializing /rw/bind-dirs with files from /var/spool/cron
[2025-02-23 09:16:21] [    4.530095] enable-firewall[482]: 2025-02-23 14:16:21 - whonix-workstation-firewall - OK: (Full torified network access allowed.)
[2025-02-23 09:16:21] [    4.539097] bind-dirs.sh[458]: Bind mounting /rw/bind-dirs/var/spool/cron onto /var/spool/cron
[2025-02-23 09:16:21] [    4.541085] bind-dirs.sh[458]: /var/lib/whonix is not a symlink
[2025-02-23 09:16:21] [    4.542240] bind-dirs.sh[458]: Initializing /rw/bind-dirs with files from /var/lib/whonix
[2025-02-23 09:16:21] [    4.550064] bind-dirs.sh[458]: Bind mounting /rw/bind-dirs/var/lib/whonix onto /var/lib/whonix
[2025-02-23 09:16:21] [    4.552782] bind-dirs.sh[458]: /var/cache/anon-base-files is not a symlink
[2025-02-23 09:16:21] [    4.555454] bind-dirs.sh[458]: Initializing /rw/bind-dirs with files from /var/cache/anon-base-files
[2025-02-23 09:16:21] [    4.557289] bind-dirs.sh[458]: Bind mounting /rw/bind-dirs/var/cache/anon-base-files onto /var/cache/anon-base-files
[2025-02-23 09:16:21] [    4.560829] bind-dirs.sh[458]: /var/lib/dummy-dependency is not a symlink
[2025-02-23 09:16:21] [    4.565297] bind-dirs.sh[458]: Initializing /rw/bind-dirs with files from /var/lib/dummy-dependency
[2025-02-23 09:16:21] [    4.571158] bind-dirs.sh[458]: Bind mounting /rw/bind-dirs/var/lib/dummy-dependency onto /var/lib/dummy-dependency
[2025-02-23 09:16:21] [    4.572259] bind-dirs.sh[458]: /var/cache/qubes-whonix is not a symlink
[2025-02-23 09:16:21] [    4.576116] bind-dirs.sh[458]: /var/cache/qubes-whonix is neither a directory nor a file and the path does not exist below /rw, skipping.
[2025-02-23 09:16:21] [    4.576192] bind-dirs.sh[458]: /var/lib/sdwdate is not a symlink
[2025-02-23 09:16:21] [    4.576627] bind-dirs.sh[458]: Initializing /rw/bind-dirs with files from /var/lib/sdwdate
[2025-02-23 09:16:21] [    4.584216] bind-dirs.sh[458]: Bind mounting /rw/bind-dirs/var/lib/sdwdate onto /var/lib/sdwdate
[2025-02-23 09:16:21] [    4.584861] auditd[662]: No plugins found, not dispatching events
[2025-02-23 09:16:21] [    4.586310] bind-dirs.sh[458]: /var/cache/whonix-setup-wizard is not a symlink
[2025-02-23 09:16:21] [    4.590180] bind-dirs.sh[458]: /var/cache/whonix-setup-wizard is neither a directory nor a file and the path does not exist below /rw, skipping.
[2025-02-23 09:16:21] [    4.593036] bind-dirs.sh[458]: /var/cache/setup-dist is not a symlink
[2025-02-23 09:16:21] [    4.593618] auditd[662]: Init complete, auditd 3.0.9 listening for events (startup state enable)
[2025-02-23 09:16:21] [    4.593916] bind-dirs.sh[458]: Initializing /rw/bind-dirs with files from /var/cache/setup-dist
[2025-02-23 09:16:21] [    4.603529] bind-dirs.sh[458]: Bind mounting /rw/bind-dirs/var/cache/setup-dist onto /var/cache/setup-dist
[2025-02-23 09:16:21] [    4.609401] bind-dirs.sh[458]: /var/lib/canary is not a symlink
[2025-02-23 09:16:21] [    4.613249] enable-firewall[482]: 2025-02-23 14:16:21 - whonix-workstation-firewall - OK: Qubes DNS firewall rules ok.
[2025-02-23 09:16:21] [    4.613857] bind-dirs.sh[458]: Initializing /rw/bind-dirs with files from /var/lib/canary
[2025-02-23 09:16:21] [    4.616493] bind-dirs.sh[458]: Bind mounting /rw/bind-dirs/var/lib/canary onto /var/lib/canary
[2025-02-23 09:16:21] [    4.620446] bind-dirs.sh[458]: /var/lib/systemcheck is not a symlink
[2025-02-23 09:16:21] [    4.622814] bind-dirs.sh[458]: Initializing /rw/bind-dirs with files from /var/lib/systemcheck
[2025-02-23 09:16:21] [    4.626586] bind-dirs.sh[458]: Bind mounting /rw/bind-dirs/var/lib/systemcheck onto /var/lib/systemcheck
[2025-02-23 09:16:21] [    4.629410] bind-dirs.sh[458]: /home/user/Downloads is not a symlink
[2025-02-23 09:16:21] [    4.632806] bind-dirs.sh[458]: Initializing /rw/bind-dirs with files from /home/user/Downloads
[2025-02-23 09:16:21] [    4.634593] bind-dirs.sh[458]: Bind mounting /rw/home/user/Downloads onto /home/user/Downloads
[2025-02-23 09:16:21] [    4.636515] bind-dirs.sh[749]: mount: /home/user/Downloads: special device /rw/home/user/Downloads does not exist.
[2025-02-23 09:16:21] [    4.636934] bind-dirs.sh[749]:        dmesg(1) may have more information after failed mount system call.
[2025-02-23 09:16:21] [    4.641909] systemd[1]: qubes-bind-dirs.service: Main process exited, code=exited, status=32/n/a
```
https://openqa.qubes-os.org/tests/129883

marmarek:
Ah, this is exactly what you've written above.

Guiiix:
This should do the job if my understanding of ``cp --parents`` is correct.

vm-systemd/bind-dirs.sh
```diff
@@ -130,6 +158,52 @@ for source_folder in /usr/lib/qubes-bind-dirs.d /etc/qubes-bind-dirs.d /rw/confi
...
+      # create resource if it does not exist
+      if ! [ -e "${path}" ] && ! [ -e "$rw_path" ]; then
+        if [ "$resource_type" = "file" ]; then
+          # for files, we need to create parent directories
+          parent_directory="$(dirname "$rw_path")"
+          echo "custom-persist: pre-creating file ${rw_path} with rights ${owner}:${group} ${mode}"
+          [ -d "$parent_directory" ] || mkdir -p "${parent_directory}"
+          touch "${rw_path}"
+        elif [ "$resource_type" = "dir" ]; then
+          echo "custom-persist: pre-creating directory ${rw_path} with rights ${owner}:${group} ${mode}"
```
marmarek, line 191: Here, and also above for file, parent directories may need owner adjustment... Imagine if somebody wants to persist /home/user/.config/some-subdir. If `/home/user/.config` doesn't exist yet, it will be created as root, likely breaking many other applications (that wouldn't be able to write to `/home/user/.config/other-subdir` for example).
I see `mkdir` has only `--mode` option, but not `--owner`... So, it may need manual loop over parent dirs (because existing dirs shouldn't have owners changed). As for what owner that should be, the same as the target file should be fine in most cases (and in the remaining corner cases, one can manually pre-create target in the template).

marmarek (changes requested):

marmarek:
But at least tests are green now :)

marmarek:
The "eject" icon there is especially confusing. Will it actually unmount the thing if you click it? Maybe for some reason Thunar thinks it's a removable device... I wonder how it gets that, there is no device attached after all (so it can't be udev, and probably not udisks either).

Guiiix:
> So I think what's happening here is that when the mounts happen in home, they should up in thunar. You tested with a file in `/var/`.

Ah yes I got the eject button too with a bind mount in ``/home``. IIUC this is the service ``gvfs-udisks2-volume-monitor`` that is responsible for listing removable devices in the sidebar. Filtering is done by looking at the mount path:

https://github.com/GNOME/gvfs/blob/master/monitor/udisks2/what-is-shown.txt

If we add ``x-gvfs-hide`` mount option in ``bind_dirs.sh`` it's better but we still see an entry in Thunar:

```
[user@fedora-41-persist ~]$ sudo mount | grep xvdb
/dev/xvdb on /rw type ext4 (rw,nosuid,nodev,relatime,seclabel,discard)
/dev/xvdb on /home/user/.bash_history type ext4 (rw,nosuid,nodev,relatime,seclabel,discard)
/dev/xvdb on /home/user/.bash_history type ext4 (rw,nosuid,nodev,relatime,seclabel,discard,x-gvfs-hide)
/dev/xvdb on /home/user/Downloads type ext4 (rw,nosuid,nodev,relatime,seclabel,discard)
/dev/xvdb on /home/user/Downloads type ext4 (rw,nosuid,nodev,relatime,seclabel,discard,x-gvfs-hide)
```
Capture d’écran_2025-02-26_21-51-25

This is related to the second problem
> Also, in the `mount` output I do notice a duplicate entry for the directory. Is this intended?
> journalctl -u qubes-bind-dirs.service -u qubes-mount-dirs.service

Duplicates mounts that might be caused by mount propagation:

https://unix.stackexchange.com/questions/464605/why-do-repeated-bind-mounts-create-entries-for-the-source-directory 

As the second entry is automatically generated, we can't add the ``x-gvfs-hide``. As for ``gvfs-udisks2-volume-monitor`` paths monitored seem to be hard-coded: 

https://github.com/GNOME/gvfs/blob/master/monitor/udisks2/gvfsudisks2volumemonitor.c#L692 

By overriding ``HOME`` environment variable in the unit file and setting its value to a something that is not an absolute path, ``g_get_home_dir()`` returns NULL and ``$HOME`` mount points are ignored. I can add the following drop-in if it's not too ugly:

```
[user@fedora-41-persist ~]$ cat /usr/lib/systemd/user/gvfs-udisks2-volume-monitor.service.d/30_qubes.conf 
[Service]
Environment="HOME=fakehome"
```
---

> Will it actually unmount the thing if you click it? 

It tries but got permission denied

vm-systemd/bind-dirs.sh
Guiiix: Should we keep 0755 mode for parent dirs?

Guiiix: ``install -d`` should do the job

marmarek:
> Duplicates mounts that might be caused by mount propagation:

This makes sense. So, it's basically caused by having `home.mount` that mounts `/home` to `/home`. Maybe instead of changing `What` field, it can be disabled some other way, for example with `ConditionPathExists=/nonexistent`?

vm-systemd/bind-dirs.sh
marmarek: > Should we keep 0755 mode for parent dirs?

should be fine (if somebody wants other, they can always create the dir manually in the template)

Guiiix:
Ah yes!! I was thinking ``ConditionPathExists`` didn't work with mount units but when I tested before, mount units were generated by ``systemd-fstab-generator`` and that was causing drop-ins to be ignored.
Now that ``home.mount`` and ``usr-local.mount`` are regular mount units (not generated ones), it works! Thank you.

So the winning combination is:
  * disable ``/home`` and ``/usr/local`` using ``ConditionPathExists``
  * mount every bind dir with ``-o x-gvfs-hide`` :)

marmarek:
A minor notes, potentially for the future:
1. In commit messages have a blank line between a summary and details.
2. Use 4 spaces for indentation, not 2 (I know the current bind-dirs.sh is quite inconsistent already).

Guiiix:
Noticed, thank you for the clarification!

vm-systemd/bind-dirs.sh
```diff
@@ -91,13 +108,19 @@ bind_dirs() {
          if [ ! -e "$fso_ro" ]; then
             ## Create empty file or directory if path exists in /rw to allow to bind mount none existing files/dirs.
             test -d "$fso_rw" && mkdir --parents "$fso_ro"
-            test -f "$fso_rw" && touch "$fso_ro"
+            if [ -f "$fso_rw" ]; then
+              parent_directory="$(dirname "$fso_ro")"
+              test -d "$parent_directory" || mkdir --parents "$parent_directory"
```
marmarek, line 113: Those two mkdirs are the ones that matter for owner/mode. But here you don't have the settings readily available... maybe you can read them `$fso_rw` parents (which is already created at this point)?

Guiiix: Ah yes my commit is a nonsense...
What about letting this part intact, using ``mk_parent_dirs`` to create ``$path`` (``$fso_ro``) parents and ``mkdir -p`` to create ``$rw_path`` (``$fso_rw``) parents here:

https://github.com/QubesOS/qubes-core-agent-linux/blob/f18831c5007190e164cece8cd1cfb9416946348b/vm-systemd/bind-dirs.sh#L200-L217

marmarek:
Thunar now don't show those dirs anymore :)

Generally I think it's all good now. I'll take a look at the whole diff one more, but should be okay.

I noticed one more potential issue - SELinux labels. new dirs have `root_t` type, instead of `user_home_t` (for directories in user home). But 1) it isn't anything new in this PR, it's how bind-dirs behave, and 2) user is running as unconfined anyway, so labels in user home don't matter that much.
This can be fixed later, if it turns out to cause issues.

marmarek (approved):

PR #555, opened 2025-02-25
fix(qubes-download-dom0-updates): manually specify vars dir in args

gianklug:
This fixes custom vars for DNF on updatevms running Fedora 41. 

Even though the custom variables are copied over [here](https://github.com/QubesOS/qubes-core-admin-linux/blob/main/dom0-updates/qubes-dom0-update#L305C37-L305C49), `dnf` only replaces them with the added args in this patch on Fedora versions > 40.

### DIFF
--- a/package-managers/qubes-download-dom0-updates.sh
+++ b/package-managers/qubes-download-dom0-updates.sh
@@ -13,6 +13,8 @@ elif [ -f "$DOM0_UPDATES_DIR/etc/yum.conf" ]; then
 fi
 # DNF uses /etc/yum.repos.d, even when --installroot is specified
 OPTS+=("--setopt=reposdir=$DOM0_UPDATES_DIR/etc/yum.repos.d")
+# DNF does not find the right vars dir automatically when --installroot is used
+OPTS+=("--setopt=varsdir=$DOM0_UPDATES_DIR/etc/yum/vars")
 CLEAN_OPTS=("${OPTS[@]}")
 # DNF verifies signatures implicitly, but yumdownloader does not.
 SIGNATURE_REGEX=""


marmarek:
> Commit signing should've been fixed, please re-run.

Looks like the bot for checking signatures doesn't support SSH signing... ignore this problem for now.

marmarek:
openQA said now installing packages for dom0 fails with:
```
Failed to download metadata (metalink: "https://yum.qubes-os.org/r41/current/host/fc41/repodata/repomd.xml.metalink") for repository "qubes-dom0-current"
```

The "r41" in the URL is wrong, should be "r4.3". Looks like it took fedora version from updatevm, instead of qubes-release version from dom0's rpmdb.
I'm not 100% sure if it's caused by this PR, but pretty likely. It worked in another test run 2 days ago. Alternative cause could be some DNF update in Fedora in the meantime, but last one I see a week ago.

marmarek:
Yes, it's this PR. A test run without it works okay.

gianklug:
Oh, well. Guess I'll have to do some debugging as to why this happens. Does the OpenQA test use any non-standard dnf variables or overwrites some standard ones? That would probably explain the errors.

marmarek:
> * system_tests_guivm_vnc_gui_interactive
>   
>   * update_guivm: [Failed](https://openqa.qubes-os.org/tests/130400#step/update_guivm/21) (test died)
>     `# Test died: command '(set -o pipefail; qubesctl --all --show-outpu...`
> 
> *

This is the thing that failed specifically. So, it's `qubes-dom0-update` call via Salt's `pkg.installed` module... Unfortunately it doesn't print full command it used...

marmarek:
You can find full salt output at https://openqa.qubes-os.org/tests/130400/file/update_guivm-qubesctl-sys-gui.log

marmarek:
No update, closing for now. Feel free to leave a comment to reopen if you are still interested in working on this.

PR #562, opened 2025-03-15
Launch `volumeicon` and `pasystray` only for audiovm

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/9158

### DIFF
--- a/autostart-dropins/pasystray.desktop
+++ b/autostart-dropins/pasystray.desktop
@@ -1,0 +1,3 @@
+[Desktop Entry]
+Exec=/usr/lib/qubes/qvm-service-wrapper audiovm pasystray
+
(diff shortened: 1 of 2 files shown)


marmarek:
It's tempting to use https://github.com/QubesOS/qubes-core-agent-linux/pull/538 instead (like in https://github.com/QubesOS/qubes-gui-agent-linux/pull/222/files), but that feature is not available in R4.2, so it will make backporting more complicated)

alimirjamali:
> It's tempting to use #538 instead (like in https://github.com/QubesOS/qubes-gui-agent-linux/pull/222/files), but that feature is not available in R4.2, so it will make backporting more complicated.

Can't we use both? (overriding the `Exec=` as well as `AutostartCondition=unless-exists` together). Shouldn't the old implementation on R4.2 ignore the latter case?

marmarek:
That would work, but on R4.3 it would check then twice. Not a bit issue, but still not very clean.

Lets keep it like this for now, and later convert it to the new method in a separate commit (that isn't going to be backported).

marmarek (approved):

PR #564, opened 2025-03-20
fix(network): no proxying of localhost connections by default

3nprob:
- fix: Set `NO_PROXY=127.0.0.1` for pacman and gentoo package managers, to avoid proxying connections to repos already accessed over localhost.
- feat: Allow overriding `ALL_PROXY` and `NO_PROXY` environment variables

Related:
- https://github.com/QubesOS/qubes-builder-archlinux/pull/76
- https://github.com/QubesOS/qubes-builderv2/pull/189


network/update-proxy-configs
```diff
@@ -155,6 +155,7 @@ fi
 if [ -e /etc/portage/make.conf ]; then
     update_conf /etc/portage/make.conf "http_proxy=\"$PROXY_ADDR\"
 https_proxy=\"$PROXY_ADDR\"
+no_proxy=\"${NO_PROXY:-127.0.0.1}\"
```
marmarek, line 158: I don't see it documented in make.conf(5) man page, do you have any source for this?

marmarek (changes requested):

marmarek:
Maybe stupid question, but you want to access a repo that is on "localhost" in the template. Why not simply configure it as `file://` instead of `http://localhost/` ?

marmarek:
But if you want to change that anyway, [tests](https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/tests/integ/vm_update.py#L496-L538) will need an update - currently they rely on accessing localhost repo via the proxy (and this is actually how it checks that proxy gets used). I wonder if anybody else relies on this feature...

marmarek:
No issue in the builder, but here it's about behavior of the final template, so the blast radius of a behavior change is bigger. What about having that qrexec proxy from updates proxy instead? It gives a bit more control to the updates proxy, but don't really need to use sys-net for it - it can be also a dedicated qube.

> Do you think it would be enough to make the tests use an alternative loopback address (`127.0.1.2` or whatever)

That may be a good idea anyway.

marmarek:
> Considering the update proxy out of the box will 403 these localhost connections

Ah, right. Then this change should be okay. Just update the tests first to use different loopback address.

marmarek:
> integration tests elsewhere to tackle?

At least here: https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/tests/integ/vm_update.py#L496-L538

marmarek:
Ok, then I guess no change is needed (will test).
As for blocking only 127.0.0.1, the only reason for that entry, is to show error page with special tag for whonix, see 6139ed59a377512418963a335f57489f1619df6a

marmarek:
> slightly OT...

There are some corner cases here indeed, but also updates proxy is allowed only for templates by default, so it's pretty limited.

marmarek (approved):

PR #570, opened 2025-04-25
SSL auth and user GPG key support for user templates repos

Guiiix:
This PR adds a wrapper in the payload sent for qubes.TemplateSearch and qubes.TemplateDownload RPC calls. This wrapper allows files to be sent to the Proxy VM that would be used by DNF .

To prevent conflict and permission issues, the files are not written in ``/etc/qubes/repo-templates/`` but in the temporary directory where the transmitted repository file is stored. Then, this repository file is patched in order to modify the path of the transmitted files.

https://github.com/QubesOS/qubes-issues/issues/9850

### DIFF
--- a/qubes-rpc/qvm-template-repo-query
+++ b/qubes-rpc/qvm-template-repo-query
@@ -26,6 +26,25 @@ repodir=$(mktemp -d)
 trap 'rm -r "$repodir"' EXIT
 cat > "$repodir/template.repo"
 
+# extract keys from wrapper in repo file
+mkdir "$repodir/keys"
+sed -i "s~/etc/qubes/repo-templates/keys/~$repodir/keys/~" "$repodir/template.repo"
+in_wrapper=false
+line_is_filename=true
+while read -r line; do
+    [[ "$line" == "###!Q!BEGIN-QUBES-WRAPPER!Q!###" ]] && in_wrapper=true && continue
+    [[ "$line" == "###!Q!END-QUBES-WRAPPER!Q!###" ]] && in_wrapper=false && continue
+    $in_wrapper || continue
+    if $line_is_filename; then
+        filename="${line:1}"
+        line_is_filename=false
+    else
+        mkdir -p "$(dirname "$filename")"
+        echo "${line:1}" | base64 -d > "$filename"
+        line_is_filename=true
+    fi
+done < "$repodir/template.repo"
+
 DNF5=false
 if [ "$(readlink /usr/bin/dnf)" = "dnf5" ]; then
     DNF5=true


qubes-rpc/qvm-template-repo-query
```diff
@@ -26,6 +26,26 @@ repodir=$(mktemp -d)
 trap 'rm -r "$repodir"' EXIT
 cat > "$repodir/template.repo"
 
+# extract keys from wrapper in repo file
+mkdir "$repodir/keys"
+cp /etc/qubes/repo-templates/keys/* "$repodir/keys/"
```
marmarek, line 31: This directory may not exist in the VM (with this feature implemented, the `qubes-repo-templates` package won't be required anymore in the VM. So, it needs to not fail in that case.

qubes-rpc/qvm-template-repo-query
```diff
@@ -26,6 +26,26 @@ repodir=$(mktemp -d)
...
+sed -i "s~/etc/qubes/repo-templates/keys/~$repodir/keys/~" "$repodir/template.repo"
+in_wrapper=false
+line_is_filename=true
+while read -r line; do
+    [[ "$line" == "###!Q!BEGIN-QUBES-WRAPPER!Q!###" ]] && in_wrapper=true && continue
+    [[ "$line" == "###!Q!END-QUBES-WRAPPER!Q!###" ]] && in_wrapper=false && continue
+    $in_wrapper || continue
+    if $line_is_filename; then
+        filename="$(echo "$line" | cut -c 2-)"
```
marmarek, line 40: Since it's bash, you can use it to avoid extra external call:
```suggestion
        filename="${line:1}"
```

qubes-rpc/qvm-template-repo-query
```diff
@@ -26,6 +26,26 @@ repodir=$(mktemp -d)
...
+        line_is_filename=false
+    else
+        mkdir -p "$(dirname "$filename")"
+        echo "$line" | cut -c 2- | base64 -d > "$filename"
```
marmarek, line 44: and similar here

marmarek (changes requested):

qubes-rpc/qvm-template-repo-query
Guiiix: Ok, so now I understand why we need to support $releasever in ``qvm-template``.

qubes-rpc/qvm-template-repo-query
Guiiix: Thank you, this looks much better

marmarek (approved):

PR #571, opened 2025-05-09
Wait for session only if necessary

ben-grande:
The 'gui' feature is not passed via QubesDB to the qube, if it was, only one service would be necessary.

For: https://github.com/QubesOS/qubes-issues/issues/1512

---

This makes the handler for `domain-start` simpler, besides not having to sleep if there is no GUI for a qube.

### DIFF
--- a/qubes-rpc/qubes.WaitForSession
+++ b/qubes-rpc/qubes.WaitForSession
@@ -1,11 +1,12 @@
 #!/bin/sh
+set -eu
 
-USERNAME="$(qubesdb-read /default-user || echo 'user')"
-
-while ! [ -e "/var/run/qubes/qrexec-server.$USERNAME.sock" ]
-do
-    sleep 0.1
-done
+if test "$(qubesdb-read /qubes-gui-enabled)" = "True"; then
+    user="$(qubesdb-read /default-user || echo 'user')"
+    while ! [ -e "/var/run/qubes/qrexec-server.$user.sock" ]; do
+        sleep 0.1
+    done
+fi
 
 systemctl --user --wait --quiet is-system-running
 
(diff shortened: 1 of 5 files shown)


qubes-rpc/qubes.WaitForSession
```diff
@@ -1,11 +1,12 @@
...
-
-USERNAME="$(qubesdb-read /default-user || echo 'user')"
-
-while ! [ -e "/var/run/qubes/qrexec-server.$USERNAME.sock" ]
-do
-    sleep 0.1
-done
+set -eu
+
+if qubesdb-read -q /qubes-gui-domain-xid >/dev/null; then
```
marmarek, line 4: That isn't a good idea, it will also happen if GUI domain is started later. Adding `-w` option (to wait for the value) would help, but would also defeat the purpose of this condition. I don't think there is an easy way to distinguish the "no gui" from the "gui domain not started yet" case from within a VM. Maybe there should be a separate qubesdb entry for this info?

marmarek (changes requested):

qubes-rpc/qubes.WaitForSession
ben-grande: > Maybe there should be a separate qubesdb entry for this info?

Yeah, didn't think of the case of GUIVM not being started yet... a separate qubesdb entry makes sense as it could cover the `guivm` property and the `gui` feature.

ben-grande: Using db entry `/qubes-gui-enabled`, now it depends on https://github.com/QubesOS/qubes-core-admin/pull/660

marmarek (approved):

PR #573, opened 2025-05-14
Disable (in updatevm) dnf ANSI coloring of output

alimirjamali:
Since dnf5 does a lot of coloring of output, it is better to disable it for UpdateVM as it messes the outputs of many actions such as: `qubes-dom0-update --action=search|list|...`

improves: https://github.com/QubesOS/qubes-issues/issues/9482


marmarek:
I have dnf5 5.2.13.1 and `dnf5 --color=never search edit` still uses color output (marks the "edit" word in green). Similarly `qubes-dom0-update --action=search edit` with this PR.

PR #576, opened 2025-05-28
updater: split qubes-download-dom0-updates.sh

piotrbartman:
Splitting the file allows for code reuse and improves readability


marmarek:
RPM package fails to build :(

PR #580, opened 2025-06-19
qubes-core-agent: remove conflict with firewalld

3nprob:
Currently, `qubes-core-agent` does not allow installation alongside `firewalld`. They seem to get along as fine as one would expect when allowed, though. Users should be allowed to use it if they want.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -79,7 +79,6 @@ Recommends:
     xterm
 Conflicts:
     qubes-core-agent-linux,
-    firewalld,
     qubes-core-vm-sysvinit,
     qubes-gui-agent (<< 4.1.6-1),
     pulseaudio-qubes (<< 4.2.0-1),
--- a/rpm_spec/core-agent.spec.in
+++ b/rpm_spec/core-agent.spec.in
@@ -125,7 +125,6 @@ Vendor:     Invisible Things Lab
 License:    GPL
 URL:        https://www.qubes-os.org
 
-Conflicts:  firewalld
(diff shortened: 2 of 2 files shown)


marmarek:
None of firewall services in qubes (qubes-firewall.service, qubes-iptables.service (which actually uses nft...), qubes-antispoof.service) work fine with firewalld. They will override rules set by the other, or at least add conflicting rules. While in some situations it may appear they "get along fine", the actual result is usually firewall not blocking things it should block, or blocking too much.

So, no, don't just remove the conflict. If you want to use firewalld, all of the mentioned services need to be adjusted to consider presence of firewalld and coordinate adding rules with them.

marmarek:
Try for example `firewall-cmd --add-port=443/tcp` - you'd expect traffic to 443 would be allowed after this command, but it won't be the case. That said, denying too much is usually safe (although may be surprising). With nft, it appears firewalld no longer removes tables/rules added by other tools (it was the case with iptables backend), so maybe it's a bit less bad nowadays.

As for other packages touching firewall, the thing about firewalld is it being commonly installed by default, and often used in various tutorials. Preventing its installation makes it obvious it won't work. If you get it running and then it doesn't work it's much more confusing behavior for users.

But since it appears it doesn't remove qubes-added rules anymore (at least not in default configuration), maybe some middle ground is to drop the conflict, but add a service drop-in to not start it by default unless explicitly enabled with qvm-service (similar to many other services)?

PR #585, opened 2025-06-21
arch, debian: Sort dependencies

3nprob:
Make the archlinux and debian specs more maintainable by consistently sorting dependency lists alphabetically.

Includes #584

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,23 +3,23 @@ Section: admin
 Priority: extra
 Maintainer: unman <unman@thirdeyesecurity.org>
 Build-Depends:
+    config-package-dev,
     debhelper,
+    desktop-file-utils,
+    dh-python,
     libpam0g-dev,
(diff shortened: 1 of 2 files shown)


debian/control
```diff
@@ -29,36 +29,36 @@ Architecture: any
...
+    psmisc,
+    python3-dbus,
     python3-gi,
+    python3-qubesdb,
     python3-xdg,
-    python3-dbus,
-    qubes-utils (>= 3.1.3),
     qubes-core-qrexec (>= 4.2.19),
     qubesdb-vm,
+    qubes-utils (>= 3.1.3),
```
marmarek, line 51: That's interesting ordering. I think I see how this happened (basically, sorted with ignoring dashes), but is it actually desirable? There are few other places here where ignoring dashes affect sort order (for example `init-system-helpers`)

marmarek (approved):

PR #593, opened 2025-07-05
fix(qubes-setup-dnat-to-ns): handle null dns dbus response

3nprob:
`resolve1.Get` dbus call is not guaranteed to return a list. This handles the case of a `None` result as equal to `[]`.

An alternative would be to fall back to `get_dns_resolv_conf` (as already done for handling known errors)

- Broken out from #592

### DIFF
--- a/network/qubes-setup-dnat-to-ns
+++ b/network/qubes-setup-dnat-to-ns
@@ -60,6 +60,9 @@ def get_dns_resolved():
         dns = resolve1.Get('org.freedesktop.resolve1.Manager',
                            'DNS',
                            dbus_interface='org.freedesktop.DBus.Properties')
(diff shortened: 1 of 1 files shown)


network/qubes-setup-dnat-to-ns
```diff
@@ -48,6 +48,7 @@ def get_dns_resolv_conf():
     return nameservers
 
 def get_dns_resolved():
+    dns = []
```
marmarek, line 51: This line isn't needed - there is no code path that leave dns unset and gets to the place where it's used. And there would be - better get an exception that is easy to find, than _silently_ break DNS.

marmarek (approved):

PR #596, opened 2025-07-05
tinyproxy: cleanup config and script

zpc0:
- Drop unused config.
- Drop no longer needed wrapper.

Additional question:
I think 60s tinyproxy timeout is too long for system updates, because there are many mirror servers to use. Should I add commit which reduces timeout to 5s or 10s?


marmarek:
Packages fail to build - make file needs to be updated.

PR #598, opened 2025-07-10
tinyproxy: hardcode /usr/bin/tinyproxy path

3nprob:
installed at same path on debian/redhat/fedora.

- follow-up to #596
- #157
- #155


CertainLach:
> Hmm, do you know what NixOS does? It surely doesn't have /usr/bin/tinyproxy... @CertainLach

Usually, everything needs to be patched for NixOS; Not many of Qubes packages even rely on PATH, and then there is lots of expectations regarding script locations

PR #603, opened 2025-09-04
Introduce a service to wait for network uplink

ben-grande:
Fixes: https://github.com/QubesOS/qubes-issues/issues/10173
For: https://github.com/QubesOS/qubes-issues/issues/1512

### DIFF
--- a/qubes-rpc/qubes.WaitForNetworkUplink
+++ b/qubes-rpc/qubes.WaitForNetworkUplink
@@ -1,0 +1,21 @@
+#!/bin/bash
+# Wait with a timeout for the uplink service to be started. Used on deferred netvm.
+set -eu
+
+# shellcheck disable=SC1091
+. /usr/lib/qubes/init/functions
+iface="$(get_qubes_managed_iface)"
+if test -z "$iface"; then
+  exit
+fi
+
+i=0
+while test $i -le 10; do
+  if systemctl is-active --quiet "qubes-network-uplink@$iface.service"; then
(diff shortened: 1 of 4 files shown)


marmarek:
> system_tests_network
> 
>     * VmNetworking_debian-13-xfce: [test_001_simple_networking_paused](https://openqa.qubes-os.org/tests/152112#step/VmNetworking_debian-13-xfce/2) (failure)
>       `~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^... AssertionError: 2 != 0`

This fails reliably on Debian 13. I don't see in logs on which side the problem is, might be related to calling as non-root, but might be also something else.

qubes-rpc/qubes.WaitForNetworkUplink
```diff
@@ -0,0 +1,20 @@
...
+. /usr/lib/qubes/init/functions
+iface="$(get_qubes_managed_iface)"
+if test -z "$iface"; then
+  exit
+fi
+
+i=0
+while test $i -le 10; do
+  if systemctl is-active --quiet "qubes-network-uplink@$iface.service"; then
+    exit 0
```
ben-grande, line 15: Check if link exists and is UP. Report with exit code if it is not.

`ip link show $iface` and grep for state.

marmarek (approved):
I suspect this might not be necessary after fixing the kernel issue mentioned in the other PR, but since it isn't fixed yet, lets have it in

PR #604, opened 2025-09-05
Add flatpak update notification and auto proxy config

alimirjamali:
Automatic proxy config for flatpak in TemplateVMs. Notify dom0 if system-wide (in TemplateVM) flatpaks are outdated.

A separate patch to `qubes-core-admin-linux` & `vmupdate` will enable easy update of (in Template) flatpaks via the updater.

related: https://github.com/QubesOS/qubes-issues/issues/2766


marmarek:
This is weird, clearly some network access uses proxy, because adding remote worked, search also works. But `flatpak remote-ls` fails with:
```
error: Unable to load summary from remote fedora: Failed to update OCI summary: While fetching https://registry.fedoraproject.org/index/static?label%3Aorg.flatpak.ref%3Aexists=1&architecture=amd64&os=linux&tag=latest: [6] Could not resolve hostname
```
But `flatpak remote-ls flathub` works. So, this is something about the fedora remote. Maybe it's because it's "oci"?
```
Name    Title           URL                                    Collection ID Subset Filter Priority Options    … … Homepage             Icon
fedora  Fedora Flatpaks oci+https://registry.fedoraproject.org -             -      -      1        system,oci … … -                    -
flathub Flathub         https://dl.flathub.org/repo/           -             -      -      1        system     … … https://flathub.org/ https://dl.flathub.org/repo/logo.svg
```

marmarek:
I haven't checked, but I'm afraid the remote-ls failure will also break updates check...

alimirjamali:
> I haven't checked, but I'm afraid the remote-ls failure will also break updates check...

Ok. It works on Fedora 43 and is broken on Fedora 42. I have to investigate

marmarek:
Converting to draft, as it needs to be merged together with the other PR that is draft too. Feel free to un-draft it when the other PR is ready.

notmuchtohide:
I tested the changes in `network/update-proxy-configs` in a Fedora 43 template with no internet connection.
I could successfully add flathub as a remote and install `org.mozilla.firefox` (in the system).
However, I could open Firefox and surf the web on my template doing `flatpak run org.mozilla.firefox`, which could be a security concern.

One way around this could be to disable `flatpak run` in templates.

notmuchtohide:
> One way around this could be to disable `flatpak run` in templates.

A better solution would be to disable network access by default in flatpak apps, using flatpak overrides. 

This approach, combined with the changes already done in `update-proxy-configs`, flatpak itself can access the internet via proxy to install, update, add remotes, etc., but flatpak apps won't have internet access.

The following configurations can be inserted in the `global` file in `XDG_DATA_HOME` (which seems to be `/home/user/.local/share/flatpak/overrides/`) so that the changes only apply to the template itself and not to app qubes based on it:

```
[Context]
shared=!network;
```

This is a contribution I would like to make to complement the solution.

marmarta:
Does this have to be in the user home dir? Can it be another XDG directory? (because it's going to be ugly to put it in templates only)

PR #639, opened 2026-03-13
don't resize volatile root

### LINKED ISSUE #10503: adding disk space to a template without running the template before running the dependent appvms breaks  appvms
[How to file a helpful issue](https://www.qubes-os.org/doc/issue-tracking/)

### Qubes OS release
I believe I have seen this on both 4.2 and 4.3

### Brief summary
adding disk space to a template without running the template before running the dependent appvms breaks  appvms

### Steps to reproduce
1. open the gui "settings" for a debian-13/debian-13-xfce template and increase the "system storage max size" by 10 gigs
2. close the settings and do *not* start the debian template
3. start a appvm that uses the template you just changed as its template

It goes for 60 seconds then aborts due to timeout, with the console messages ending in:

...
.[K[ .[0;31m*.[0;1;31m*.[0m.[0;31m*  .[0m] Job qubes-rootfs-resize.service/start running (55s / no limit)
.[K[.[0;31m*.[0;1;31m*.[0m.[0;31m*   .[0m] Job qubes-rootfs-resize.service/start running (55s / no limit)
.[K[.[0;1;31m*.[0m.[0;31m*    .[0m] Job qubes-rootfs-resize.service/start running (56s / no limit)
[root@dom0 qubes-rpc]# xl list

Now, lets keep going!

4. start the templatevm that we have been dealing with
    (it should resize, then start the application you asked it to run within a few seconds)
6. once it has started, shut it down again
7. start the same appvm as before
This time it should work!

### Expected behavior
starting without a timeout error

### Actual behavior
a timeout error

### Additional information

I'm using nvmes, so i dont really feel this is a performance issue

I've seen this multiple times.  

Since the solution is to just run the template and shut it down, this probably isn't that critical of a priority.

hippalectryon-0:
Fixes https://github.com/QubesOS/qubes-issues/issues/10503

If we resized a TemplateVM but never booted it, then boot an AppVM that depends on the template, `resize-rootfs-if-needed.sh` will resize the volatile root volume through CoW, which is very slow and likely unexpected.

This skips resizing root volumes for AppVMs and DispVMs.

Note: I'm not sure this is the best way to do that and would appreciate feedback - for instance if there was away to explicitly check whether the root volume is volatile rather than hardcoding VM classes..

Note: from an User Experience pt of view this is still iffy - less technical users will not expect to have to run the TemplateVM once to apply the resizing changes - but it's a separate issue.

### DIFF
--- a/init/resize-rootfs-if-needed.sh
+++ b/init/resize-rootfs-if-needed.sh
@@ -11,6 +11,14 @@ if [ "$(blockdev --getro /dev/xvda)" -eq "1" ]; then
     exit 0
 fi
 
+# if root volume is a volatile snapshot (AppVM/DispVM), skip resize —
+# it must happen in the TemplateVM where changes persist
+vm_type=$(qubesdb-read /qubes-vm-type 2>/dev/null || true)
+if [ "$vm_type" = "AppVM" ] || [ "$vm_type" = "DispVM" ]; then
+    echo "volatile root volume ($vm_type), skipping resize" >&2
+    exit 0
+fi
+
 sysfs_xvda="/sys/class/block/xvda"
 
 # if root filesystem is already using (almost) the whole disk


marmarek:
> for instance if there was away to explicitly check whether the root volume is volatile

Check for `/run/qubes/persistent-full`? It's created based on persistence announced in qubesdb: https://doc.qubes-os.org/en/latest/developer/debugging/vm-interface.html#keys-exposed-by-dom0-to-vm

> Note: from an User Experience pt of view this is still iffy - less technical users will not expect to have to run the TemplateVM once to apply the resizing changes - but it's a separate issue.

Yes... I understand why skipping may be useful sometimes, but it creates inconsistency. For example if user wants just bigger /var/tmp, it isn't obvious that they need to also start the template after such change. Maybe at least add a message in GUI settings that resize will be applied only the next time the VM is started (if it isn't running at that time)? Another, rather obscure idea, would be to mark it as updates available, which also kinda prompts for starting such template, but it's ugly (especially since we don't have more info about available updates, so can't put real reason there).

PR #642, opened 2026-03-20
Fix for #10797: qvm-open-in-vm should now work in KDE-based cubes

### LINKED ISSUE #10797: qvm-open-in-dvm/qvm-open-in-vm doesn't work with KDE-based templates
### Qubes OS release

R4.3, but might affect older versions

### Brief summary

`qvm-open-in-dvm` and `qvm-open-in-vm` do not work with DispVM based on KDE-based templates.

### Steps to reproduce

1. Create a KDE-based template (for example, following [this guide](https://forum.qubes-os.org/t/installing-kde-in-a-debian-13-template/37324)),
2. Create a DisposableVM Template based on that template. Set it as default dispvm (either globally, or for a VM you will use in the next step)
3. Start any VM, create a non-empty txt file there.
4. Call `qvm-open-in-dvm` on this file

### Expected behavior

File opens in a new DispVM.
Changes to the file are transferred back to the source VM (after closing the editor).

### Actual behavior

The DispVM starts and then immediately shuts down.

### Additional information

Similar to https://github.com/QubesOS/qubes-issues/issues/6884

What's happening here is that `qubes-open` is using `xdg-open`, which in turn uses `kde-open` or `kde-open5` in the KDE-based cube. These (`kde-open`/`kde-open5`) start the application that handles the file in a way that it is not their child process. 

This is how `pstree` looks when opening a file in a working DispVM (where `firefox-esr` happens to be the application handling the file type):

```
        ├─qrexec-fork-ser───qrexec-fork-ser─┬─logger
        │                                   └─vm-file-editor───qubes-open─┬─cat
        │                                                                 └─xdg-open───firefox-esr
```

And this is how it looks in a cube where `kde-open`/`kde-open5` was used to launch the application (in this case, `gwenview`):

```
        ├─systemd─┬─(sd-pam)
        │         ├─(...)
        │         ├─gwenview───13*[{gwenview}]
```

`qubes-open` is expected to wait until all its child processes are done; after it exits, the parent proces -- `vm-file-editor` -- cleans up the temporary directory with the file being opened/edited, and exits, allowing the cube to shut down.

Since `kde-open`/`kde-open5` immediately return, and since the application process is not a child of `qubes-open`, the DispVM immediately shuts down.

In fact, this also affects `qvm-open-in-vm` when the target cube uses a KDE-based template. In that case, the application is started, but by the time it actually fully starts, the file is gone, as `vm-file-editor` already managed to delete it after `qubes-open` exited.

### Proposed solution

Currently `qubes-open` [uses a combination of `exec` and `cat`](https://github.com/QubesOS/qubes-core-agent-linux/blob/8214ce1cbad1525153de70342bc163fa5f6f07aa/qubes-rpc/qubes-open#L11) to keep stdout/stderr connected even if the handler (say, `exo-open`):

```bash
# The "| cat" is rather silly hack, to wait for all the children processes too
# (unless they close their stdout). This is needed on Xfce template, because
# exo-open starts the application in the background, but keeps the original
# stdout/err connected.
export DE=generic
exec xdg-open "$@" | cat
```

This does not work for `kde-open`/`kde-open5`. Instead, I tested a quick-fix based on `pidwait`:

```bash
export DE=generic
xdg-open "$@"
pidwait -f "$1"
```

This works well in KDE-based templates, for viewing as well as editing files. Testing in GNOME-based and Xfce-based templates is my next step.

I'm sure this is not a perfect solution, there is possibly a race condition here where some other application *could* open that same temporary file and get picked up by `pidwait`. This seems unlikely to me, but flagging just as well. That said, the `| cat` method used currently is also described as a "silly hack".

rysiekpl:
`kde-open`/`kde-open5` start the process that handles the file in a way that the old `| cat` hack does not work.

This fixes it. Tested on GNOME-, Xfce-, and KDE-based templates.

Closes QubesOS/qubes-issues#10797

### DIFF
--- a/qubes-rpc/qubes-open
+++ b/qubes-rpc/qubes-open
@@ -8,9 +8,20 @@ fi
 # gvfs-open, kde-open, and possibly others don't wait for editor to be
 # closed, which is critical behaviour for DisposableVM (which gets destroyed
 # after this process exits).
-# The "| cat" is rather silly hack, to wait for all the children processes too
-# (unless they close their stdout). This is needed on Xfce template, because
-# exo-open starts the application in the background, but keeps the original
-# stdout/err connected.
+#
+# The `pidwait -A -f "$1"` is rather silly hack, to wait for the process that is
+# passed the temporary file path as a parameter, but ignoring parent processes.
+#
+# This is needed on Xfce template, because exo-open starts the application in
+# the background; and in templates built around KDE, because they start the
+# file handling process in a way that it is not a child of the shell that
+# spawned this script.
+#
+# We also want to make sure that `xdg-open`'s exit code is emitted back to
+# parent process in case of any errors.
 export DE=generic
-exec xdg-open "$@" | cat
+xdg-open "$@"
+XDG_OPEN_EXIT_CODE=$?
+pidwait -A -f "$1"
+exit $XDG_OPEN_EXIT_CODE
+


Atrate:
```
# The `pidwait -A -f "$1"` is rather silly hack
```

Is it still a "hack"? I get describing `| cat` like that, but isn't this a quite normal approach to waiting for a process?

marmarek (approved):

PR #646, opened 2026-03-27
qubes.GetImageRGBA: fix SVG conversion for images without explicit dimensions

### LINKED ISSUE #9145: qvm-convert-img fails to convert SVG image
### Qubes OS release

R4.2

### Brief summary

Try to convert 

### Steps to reproduce

Download [Whonix-logo.svg](https://www.whonix.org/w/images/1/19/Whonix-logo.svg).
```sh
curl -O https://www.whonix.org/w/images/1/19/Whonix-logo.svg
```

Try to convert image:
```sh
qvm-convert-img Whonix-logo.svg Whonix-logo.safe.svg
```

### Expected behavior

Command `qvm-convert-img` converts the image.

### Actual behavior

```sh
gm identify: No decode delegate for this image format (/tmp/qimg-7C8BP9Rv.png).
gm identify: Request did not return an image.
```
```python3
  File "/usr/lib/qubes/qimg-convert-client", line 50, in <module>
    main()
  File "/usr/lib/qubes/qimg-convert-client", line 44, in main
    img = qubesimgconverter.Image.get_through_dvm(args.src)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/qubesimgconverter/__init__.py", line 320, in get_through_dvm
    return cls.get_from_stream(sys.stdin.buffer, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/qubesimgconverter/__init__.py", line 197, in get_from_stream
    raise ValueError('No icon received')
ValueError: No icon received
```

Converting the [qubes-logo-home.svg](https://www.qubes-os.org/attachment/icons/qubes-logo-home.svg) works while the [Whonix-logo.svg](https://www.whonix.org/w/images/1/19/Whonix-logo.svg) fails.

Jayant-kernel:
SVGs that specify dimensions only via viewBox (no explicit pixel width/height attributes) cause `gm identify` to return 0 for both width and height. Passing `-w 0 -h 0` to `rsvg-convert` produces an empty or corrupt PNG, which then causes the second `gm identify` to fail with "No decode delegate for this image format".

When `gm identify` returns zero or empty dimensions for an SVG, let `rsvg-convert` determine the natural size from the viewBox instead. The actual dimensions are then read from the resulting PNG by the existing `gm identify` call that follows.

Converting the Qubes logo (which has explicit pixel dimensions) is unaffected.

Fixes: https://github.com/QubesOS/qubes-issues/issues/9145

### DIFF
--- a/qubes-rpc/qubes.GetImageRGBA
+++ b/qubes-rpc/qubes.GetImageRGBA
@@ -21,22 +21,29 @@ elif ! [ -r "${filename}" ]; then
     exit 1
 fi
 
-s="$(gm identify -format '%w %h %m' "$filename")"
-w="$(echo "$s"|cut -d " " -f 1)"
-h="$(echo "$s"|cut -d " " -f 2)"
-m="$(echo "$s"|cut -d " " -f 3)"
+read -r w h m << EOF
+$(gm identify -format '%w %h %m' "$filename")
+EOF
 if [ "$m" = SVG ]; then
     tmpfile2="$(mktemp /tmp/qimg-XXXXXXXX.png)"
-    rsvg-convert -w "$w" -h "$h" -o "$tmpfile2" "$filename"
+    if [ -n "$w" ] && [ "$w" -gt 0 ] && [ -n "$h" ] && [ "$h" -gt 0 ]; then
+        rsvg-convert -w "$w" -h "$h" -o "$tmpfile2" "$filename"
+    else
+        rsvg-convert -o "$tmpfile2" "$filename"
+        # re-read dimensions from the rendered PNG since SVG had no explicit size
+        read -r w h << EOF
+$(gm identify -format '%w %h' "$tmpfile2")
+EOF
+    fi
     # downscale the image if necessary
     if [ -n "$forcemaxsize" ] && \
             { [ "$w" -gt "$forcemaxsize" ] || [ "$h" -gt "$forcemaxsize" ]; }; then
         gm convert "$tmpfile2" -scale "${forcemaxsize}x${forcemaxsize}" "$tmpfile2"
     fi
     # read the size again, because icon may not be a square or could have changed with convert
-    s="$(gm identify -format '%w %h' "$tmpfile2")"
-    w="$(echo "$s"|cut -d " " -f 1)"
-    h="$(echo "$s"|cut -d " " -f 2)"
+    read -r w h << EOF
+$(gm identify -format '%w %h' "$tmpfile2")
+EOF
     filename="$tmpfile2"
 fi
 echo "$w $h"
--- a/qubesagent/test_getimgrgba.py
+++ b/qubesagent/test_getimgrgba.py
@@ -1,0 +1,113 @@
+#!/usr/bin/env python3
+# The Qubes OS Project, https://www.qubes-os.org/
+#
+# Copyright (C) 2026 Jayant Saxena <jayantmcom@gmail.com>
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""Unit tests for qubes.GetImageRGBA script.
+
+Tests the fix for QubesOS/qubes-issues#9145: SVG images without explicit
+width/height attributes should be converted successfully.
+"""
+
+import os
+import subprocess
+import tempfile
+import unittest
+
+SCRIPT_PATH = os.path.join(
+    os.path.dirname(__file__), '../qubes-rpc/qubes.GetImageRGBA')
+
+# Minimal SVG with explicit width/height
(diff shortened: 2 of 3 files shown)


marmarek (changes requested):
This is missing the max size enforcement just below the `rsvg-convert` call.

qubes-rpc/qubes.GetImageRGBA
```diff
@@ -27,7 +27,15 @@ h="$(echo "$s"|cut -d " " -f 2)"
...
 if [ "$m" = SVG ]; then
     tmpfile2="$(mktemp /tmp/qimg-XXXXXXXX.png)"
-    rsvg-convert -w "$w" -h "$h" -o "$tmpfile2" "$filename"
+    if [ -n "$w" ] && [ "$w" -gt 0 ] && [ -n "$h" ] && [ "$h" -gt 0 ]; then
+        rsvg-convert -w "$w" -h "$h" -o "$tmpfile2" "$filename"
+    else
+        rsvg-convert -o "$tmpfile2" "$filename"
+        # re-read dimensions from the rendered PNG since SVG had no explicit size
+        s="$(gm identify -format '%w %h' "$tmpfile2")"
+        w="$(echo "$s"|cut -d " " -f 1)"
```
ben-grande: I don't like echo because it may interpret the variable as an option. It is also unnecessary if you assign the variable with `read`.

Can you do:

```sh
gm identify -format '%w %h' -- "$tmpfile2" | read -r w h
```

Jayant-kernel: Replaced echo "$s"|cut with gm identify ... | read -r w h throughout as you  suggested.

marmarek: Does it actually work? I don't think you can use `read` this way in a pipeline (effectively in a subshell)...
```
[user@disp8053 ~]$ echo a b c | read a1 a2 a3
[user@disp8053 ~]$ echo "value: $a1"
value: 
```

ben-grande: ... I apparently tested this only on Zsh. Indeed it doesn't seem to work on (Ba)sh.

ben-grande: Abandon the `read` and use `printf '%s'` instead of `echo`.

ben-grande: Also, test what I say, don't just write it.

ben-grande: > Abandon the `read` and use `printf '%s'` instead of `echo`.

Or use variable expansion.

```
% s="a b"
% w="${v% *}"
% h="${s##* }"
```

Jayant-kernel:
I have replaced `cmd | read -r var` with a heredoc pattern (`read -r var << EOF ... EOF`)
 so that `read` runs in the current shell and the variables are actually set.

marmarek:
This breaks retrieving app icons from templates:
```
[2026-04-05 08:07:31] [   81.423740] qubes.GetImageRGBA+-sys-gui[1620]: [0;1;39mgm identify: Unrecognized option (--).[0m
[2026-04-05 08:07:31] [   81.499271] qubes.GetImageRGBA+-sys-gui[1620]: [0;1;39mgm convert: Option '-size' requires an argument or argument is malformed.[0m
```

Jayant-kernel:
I have removed -- from the gm identify calls — gm doesn't support it, and since $tmpfile2 always starts with /tmp/ it won't be mistaken for a flag. 
Should I  fix the "Unrecognized option (--) error.

ben-grande:
@Jayant-kernel Have you tried with [Whonix-logo.svg](https://www.whonix.org/w/images/1/19/Whonix-logo.svg) image described on the issue: https://github.com/QubesOS/qubes-issues/issues/9145?

Jayant-kernel:
@ben-grande 
Removed -- from the gm identify calls and signed the commit. Should be good now.

ben-grande:
> @Jayant-kernel Have you tried with [Whonix-logo.svg](https://www.whonix.org/w/images/1/19/Whonix-logo.svg) image described on the issue: [QubesOS/qubes-issues#9145](https://github.com/QubesOS/qubes-issues/issues/9145)?

Please add integration test to know if it was fixed. Maybe the test should be in [qubes-app-linux-img-converter](https://github.com/QubesOS/qubes-app-linux-img-converter), but you may need to adjust the directory structure there to be a proper python package.

Jayant-kernel:
@ben-grande I've added an integration test in a separate PR:

I have added an integration test in QubesOS/qubes-app-linux-img-converter#25 - tests SVG without explicit dimensions and also sets up the directory as a proper Python package.

marmarek:
> Please add integration test to know if it was fixed. Maybe the test should be in [qubes-app-linux-img-converter](https://github.com/QubesOS/qubes-app-linux-img-converter),

The script doesn't interact with many system components, so IMO unit test would suffice. And unit test doesn't need to be installed in dom0, can be in this repo. This would avoid the need to introduce all the boilerplate needed for integration tests (which is currently missing in the test PR).

Jayant-kernel:
@marmarek 
Added unit test in qubesagent/test_getimgrgba.py covering in this =
- SVG with explicit dimensions (normal case)
- SVG without dimensions, viewBox only (regression for #9145)
- Non-existent file

Jayant-kernel:
@marmarek 
I've added a unit test for the SVG fix . It covers the case where an SVG has only a viewBox and no explicit width/height, which is what issue #9145 was about.
The test follows the same pattern as the other tests in qubesagent/ so it should get picked up automatically by the test runner.

ben-grande:
See CI rpm build errors.

Jayant-kernel:
CI rpm build errors.

solved

marmarek (approved):

PR #648, opened 2026-04-16
Add extrepo with offline data support for Debian templates

### LINKED ISSUE #9758: Document extrepo for installing software not in Debian repositories
### The problem you're addressing (if any)

Installing third-party programs (not in distro repos) like Signal and Element can be difficult.  Official installation instructions don’t work in templates because they require Internet access.

### The solution you'd like

Debian provides the extrepo mechanism for installing software not in Debian.  Together with extrepo-offline-data, it can be used to safely and securely install Element Desktop, Signal Desktop, and lots of other software.

Qubes OS should provide it by default with a reasonable configuration:

- `extrepo` and `extrepo-offline-data` packages installed.
- No external repos enabled.
- `extrepo` pointed at `file:///usr/share/extrepo/offline-data` by default:
  ```yaml
  ---
  url: file:///usr/share/extrepo/offline-data
  dist: debian
  version: bookworm
  # To enable repositories that host software with non-DFSG-free licenses,
  # uncomment "contrib" and/or "non-free" in the list below.
  enabled_policies:
  - main
  # - contrib
  # - non-free
  ```

Furthermore, its use should be documented, if only by linking to the Debian documentation.

### The value to a user and who that user might be

Users will safely be able to install third-party software.

### Completion criteria checklist

_No response_

nihalxkumar:
A Qubes-specific extrepo configuration for Debian templates, using `config-package-dev` displacement so it does not conflict with the `extrepo` package's own `/etc/extrepo/config.yaml`.

The installed config points extrepo at the local offline data instead of fetching from the network. 

This allows users to safely enable external repositories (e.g. Signal, Element) in templates without requiring direct network access or manually managing GPG keys. All three policies (`main`, `contrib`, `non-free`) are enabled.

The config is only installed for Debian, since extrepo is a Debian-specific tool.

closes: https://github.com/QubesOS/qubes-issues/issues/9758

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -62,6 +62,8 @@ Depends:
     ${misc:Depends},
 Recommends:
     cups,
+    extrepo,
+    extrepo-offline-data,
     gnome-terminal,
     gnome-themes-standard,
     haveged,
--- a/debian/control
+++ b/debian/control
@@ -62,8 +62,6 @@ Depends:
     ${misc:Depends},
 Recommends:
     cups,
-    extrepo,
-    extrepo-offline-data,
     gnome-terminal,
     gnome-themes-standard,
     haveged,
@@ -80,11 +78,14 @@ Recommends:
     xserver-xorg-core,
     xsettingsd,
     xterm,
+Provides:
+    ${diverted-files},
 Conflicts:
     pulseaudio-qubes (<< 4.2.0-1),
     qubes-core-agent-linux,
     qubes-core-vm-sysvinit,
     qubes-gui-agent (<< 4.1.6-1),
+    ${diverted-files},
 Description: Qubes core agent
  This package includes various daemons necessary for qubes domU support,
  such as qrexec services.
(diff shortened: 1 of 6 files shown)


marmarek:
I have two main remarks:
1. The `Recommends` for just some other useful software should be in [meta-packages](https://github.com/QubesOS/qubes-meta-packages/blob/main/debian/control), not here.
2. While the extrepo config might be here (or a separate `qubec-extrepo` sub-package?), it fails to install this way due to file conflict - see CI logs. You'll need to use displace feature of  [config-package-dev](https://github.com/QubesOS/qubes-meta-packages/blob/main/debian/control) to do that.

nihalxkumar:
@marmarek  I have addressed both points. If you have any other suggestions, please let me know.
Apologies for the delayed reply; I was traveling and am now back with full force.

marmarek:
> The `Recommends` for just some other useful software should be in [meta-packages](https://github.com/QubesOS/qubes-meta-packages/blob/main/debian/control), not here.

~This is still here.~ No, it's not.

marmarek (approved):

PR #660, opened 2026-07-09
Remove PackageKit's ProxyHTTP (dead code)

notmuchtohide:
The removed code has no effect providing proxy in PackageKit as it seems ProxyHTTP is no longer an option in PackageKit.conf since PackageKit v 0.8.14 (https://github.com/PackageKit/PackageKit/commit/3afcfcd).

Tested in clean Fedora-43 (PackageKit-1.3.6-1) and Debian-13 (packagekit 1.3.1-1). It was confirmed that removing this code had no effect on installing packages from pkcon or gnome-software.


marmarta (approved):

marmarta:
(ignore codecov red cross, it's being unreasonable)

PR #661, opened 2026-07-10
Handle zramctl command failure gracefully

Rudd-O:
Elide unnecessary failure when no zram devices are setup.

### DIFF
--- a/vm-systemd/setup-minimal-vm
+++ b/vm-systemd/setup-minimal-vm
@@ -37,8 +37,8 @@ disable_zram() {
     systemctl mask --runtime --now systemd-zram-setup@zram0
     ln -nsf /dev/null /run/systemd/zram-generator.conf
     systemctl daemon-reload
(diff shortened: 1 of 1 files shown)


vm-systemd/setup-minimal-vm
```diff
@@ -38,7 +38,7 @@ disable_zram() {
     ln -nsf /dev/null /run/systemd/zram-generator.conf
     systemctl daemon-reload
     if command -v zramctl >/dev/null; then
-        zramctl -r zram0
+        zramctl -r zram0 || true # Elide unnecessary failure when no zram devices are setup.
```
marmarek: What about checking if `/dev/zram0` exists first?

marmarek (changes requested):

Rudd-O:
That would be a good idea.  I am not at a console logged into GH to do this right now, and won't be for several days.

marmarek (approved):

PR #666, opened 2026-08-07
arch: install Thunar actions with a pacman hook

### LINKED ISSUE #9740: Archlinux: thunar extension doesn't work
## Observation

openQA test in scenario qubesos-4.3-templates-x86_64-system_tests_gui_interactive@64bit fails in
[gui_filecopy](https://openqa.qubes-os.org/tests/126915/modules/gui_filecopy/steps/16)

Image

The "copy to other qube" etc entries are not there.

## Reproducible

Fails since (at least) Build [2025012605-4.3-archlinux](https://openqa.qubes-os.org/tests/126883)

## Expected result

Last good: [2025012522-4.3-archlinux](https://openqa.qubes-os.org/tests/126796) (or more recent)

## Further details

Always latest result in this scenario: [latest](https://openqa.qubes-os.org/tests/latest?arch=x86_64&distri=qubesos&flavor=templates&machine=64bit&test=system_tests_gui_interactive&version=4.3)

besendorf:
Fixes QubesOS/qubes-issues#9740

Arch Linux installs the Qubes Thunar actions under `/usr/lib/qubes/uca_qubes.xml`, but Thunar only loads system custom actions from `/etc/xdg/Thunar/uca.xml`. Add a package-specific pacman post-transaction hook that appends the Qubes actions after `qubes-vm-thunar` is installed or upgraded, matching the existing Debian and Fedora packaging behavior.

This pull request was prepared with AI assistance.

### DIFF
--- a/archlinux/PKGBUILD.qubes-update-thunar
+++ b/archlinux/PKGBUILD.qubes-update-thunar
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+uca_file=/etc/xdg/Thunar/uca.xml
+
+if [ -f "$uca_file" ]; then
+    cp -p "$uca_file" "$uca_file.bak"
+    #shellcheck disable=SC2016
+    sed -i '$e cat /usr/lib/qubes/uca_qubes.xml' "$uca_file"
+fi
--- a/archlinux/PKGBUILD.qubes-update-thunar
+++ b/archlinux/PKGBUILD.qubes-update-thunar
@@ -4,7 +4,8 @@ set -e
 
 uca_file=/etc/xdg/Thunar/uca.xml
 
-if [ -f "$uca_file" ]; then
+if [ -f "$uca_file" ] && \
+        ! grep -qF '<unique-id>1507455450991127-4</unique-id>' "$uca_file"; then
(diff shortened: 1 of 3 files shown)


marmarek:
After actually installing the qubes-vm-thunar package, this appears to be working now: https://openqa.qubes-os.org/tests/193118#step/gui_filecopy/24

archlinux/PKGBUILD.qubes-update-thunar
```diff
@@ -0,0 +1,11 @@
...
+
+set -e
+
+uca_file=/etc/xdg/Thunar/uca.xml
+
+if [ -f "$uca_file" ]; then
+    cp -p "$uca_file" "$uca_file.bak"
+    #shellcheck disable=SC2016
+    sed -i '$e cat /usr/lib/qubes/uca_qubes.xml' "$uca_file"
+fi
```
marmarek: Since this is running on package update, it should avoid adding the same actions multiple times. The easiest way would be checking if .bak file already exists, but alternatively, check if some of the action is already included in the file (via `grep`?)

marmarek (changes requested):

marmarek (approved):

PR #668, opened 2026-08-20
Gnome Software with proxy in templates

notmuchtohide:
This is needed to make flatpak work in Gnome Software.

Code inspired by [alimirjamali](https://github.com/alimirjamali/qubes-core-agent-linux/blob/659e8e620ac8d0191f65a68db2058e4c6ee2a09d/network/update-proxy-configs).


network/update-proxy-configs
```diff
@@ -186,3 +186,20 @@ RSYNC_PROXY=\"${PROXY_ADDR_BASE}\""
         update_conf /etc/hosts ""
     fi
 fi
+
+
+# Gnome Software
+if [ -f /usr/bin/gnome-software ]; then
+    if [ -n "$PROXY_ADDR" ]; then
+        mkdir -p /usr/lib/systemd/user/gnome-software.service.d
```
marmarek: Dynamic drop-ins beter create in `/run/systemd/user`

marmarek (changes requested):
Given security concerns, I think this should be additionally conditional on the plugin presence (https://github.com/notmuchtohide/qubes-gnome-software-plugin/pull/1)

marmarek (approved):

PR #675, opened 2026-09-03
qfile-unpacker: fix concurrent mount/unmount race via flock (#5950)

### LINKED ISSUE #5950: Copy simultaniously: "file-unpacker: Fatal error: Cannot umount incoming directory (error type: Invalid argument)"
**Qubes OS version**
R4.0.
The dom0 is up to date from default repo on the date 2020.07.13. 

**Affected component(s) or functionality**
qvm-copy / qfile-unpacker

**Brief summary**
When copying files simultaneously - the message with an error pops up in target vm: "_qfile-unpacker: Fatal error: Cannot umount incoming directory (error type: Invalid argument)_" after last copy process finishes.

**To Reproduce**
1. Copy 5 files of ~300 MiB each from source-vm to target-vm via GUI commands.
2. Copy other 5 files of ~300 MiB each from source-vm to target-vm via GUI commands while the first copy is still ongoing.
Reproducible in 100% cases.

**Expected behavior**
Copy processes finished without any error messages.

**Actual behavior**
The first copy process finishes with no issues. The second process finishes and pops this window in target-vm with a color of target-vm:
> Caption: [target-vm] Error
> Text:    qfile-unpacker: Fatal error: Cannot umount incoming directory (error type: Invalid argument)
> Buttons: OK

**Additional context**
1. Probably it's the same with qvm-move as it's almost the same code plus remove of original files.
2. Probably both process some kind of "mount" the target directory QubesIncoming in target-vm and the first process to finish "umounts" it and for the second it's already "umount"-ed when it finished. Not sure what "mount/umount" means in this context, though.

**Solutions you've tried**
Copy files without simultaneous copy processes for the same target-vm.

**Related issues**
I searched for "qfile-unpacker" issues but did not find any with the same exact problem.

datenbar:
Fixes QubesOS/qubes-issues/issues/5950

## Problem

When two `qvm-copy` operations to the same target VM run concurrently,
each invocation of `qfile-unpacker` independently bind-mounts and later
unmounts `incoming_dir`. Since both mounts stack at the same path in
the same mount namespace, `umount2()` can remove the wrong process's mount layer, causing:

  `  qfile-unpacker: Fatal error: Cannot umount incoming directory
    (error type: Invalid argument) `

## Fix

Instead of every invocation independently mounting and unmounting,
coordinate via flock() on incoming_dir itself :

- Each worker holds a LOCK_SH on incoming_dir for its entire lifetime.
- On entry: attempt LOCK_EX (non-blocking). Winner checks whether
  already mounted, mounts if not, downgrades to LOCK_SH. Loser blocks
  on LOCK_SH (waiting for the current holder's mount decision), then
  re-checks mount state, if still unmounted then releases SH and retries EX.
- On exit: release LOCK_SH, blocking-acquire LOCK_EX, check
  mount state, unmount only if still mounted.

- Crash safety comes from flock() itself since kernel cleans up after a process dies.

- mount()/umount2() use the resolved absolute path throughout, since a process's "." can end up resolving through an lazily-detached mount reference.

## Testing

Validated qfile-unpacker.c linked
against minimal local stubs of gui-fatal.h/libqubes-rpc-filecopy.h
(do_unpack_ext replaced with a sleep to simulate copy duration, no
change to any mount/lock logic under test).

Tested with N=64 concurrent qfile-unpacker, invocations and numerous iterations with varied timings:

- Normal timing: clean, 0 failures.
- --copy-us 0 (max mount/unmount churn): 0 incorrect outcomes. Some
  iterations show >1 mount/unmount cycle per batch (late joiner races a
  teardown) expected self-healing.
- Random SIGKILL fault injection, incl. pre-mount-completion kills:
  recovers correctly every time.

## Important Notes 
The mount-status check uses statx(STATX_MNT_ID), which requires kernel
5.8+. I haven't been able to confirm Qubes' actual minimum supported
kernel floor for VMs/templates.

## AI disclosure

Used Claude among some other models to discuss, research and write tests against my proposed fixes and review the final file before the files were reviewed and verified by me.

### DIFF
--- a/qubes-rpc/qfile-unpacker.c
+++ b/qubes-rpc/qfile-unpacker.c
@@ -1,9 +1,14 @@
+#include <asm-generic/errno.h>
 #define _GNU_SOURCE
 #include <grp.h>
 #include <unistd.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <pwd.h>
+#include <sys/file.h>
+#include <libgen.h>
+#include <errno.h>
+#include <stdbool.h>
 #include <sys/stat.h>
 #include <sys/mount.h>
 #include <sys/wait.h>
@@ -93,6 +98,22 @@ uid_t parse_uid(const char *user)
             errno != 0 || *end != '\0' || uid != u)
         gui_fatal("Invalid user ID argument");
     return uid;
+}
+
+bool is_mounted(const char *abs_path, int parent_fd)
+{
+    struct statx stx_self, stx_parent;
+
+    if (statx(AT_FDCWD, abs_path, 0, STATX_MNT_ID, &stx_self))
+      gui_fatal("statx for directory failed");
+    
+    if (statx(parent_fd, "", AT_EMPTY_PATH, STATX_MNT_ID, &stx_parent))
+      gui_fatal("statx for parent directory failed");
+
+    if (!(stx_self.stx_mask & STATX_MNT_ID))
+        gui_fatal("STATX_MNT_ID not populated, kernel may not support it");
+
+    return (stx_self.stx_mnt_id != stx_parent.stx_mnt_id);
 }
 
 int main(int argc, char ** argv)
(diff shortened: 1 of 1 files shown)


datenbar:
CI failed due to three runners getting StoragePoolExceptions

marmarek:
Tests are still in progress, but I already see there are several failures...
```
[2026-09-23 13:47:00] [  191.858202] qubes.Filecopy+-work[1746]: [0;1;39mqfile-unpacker: Fatal error: Could not acquire lock on directory /home/user/QubesIncoming/work (error type: Permission denied)[0m
```

datenbar:
That's interesting, flock doesn't have EACCES in the return error codes.  Could you give me the link of the full log?

marmarek:
It's in `var/log/xen/console/guest-personal.log` in https://openqa.qubes-os.org/tests/197327/file/gui_filecopy-var_log.tar.gz
Unfortunately in most cases this message ends up in `~/.xsession-errors` which isn't saved in tests...
But also, it seems to affect mostly Fedora (see https://openqa.qubes-os.org/tests/197320), so maybe it's related to SELinux?

datenbar:
It seems to be SELinux since personal boots with selinux=1 but there are no SELinux logs in the tar. Is it possible to somehow acquire the logs of personal's own audit subsystem around the failure timestamp? It seems that personal's audit log for this run wasn't captured anywhere I can find.

marmarek:
Something like this:
```
type=AVC msg=audit(1790287753.657:132): avc:  denied  { lock } for  pid=1105 comm="qfile-unpacker" path="/home/user/QubesIncoming/sys-firewall" dev="xvdb" ino=124 scontext=unconfined_u:unconfined_r:qubes_qfile_unpacker_t:s0-s0:c0.c1023 tcontext=unconfined_u:object_r:qubes_unpacked_file_t:s0 tclass=dir permissive=0
```

PR #677, opened 2026-09-14
mount-dirs: bail out custom-persist for TemplateVMs

### LINKED ISSUE #11054: Custom-persist in template creates `/home` and `/usr/local` on 'root' volume
### Qubes OS release

Qubes OS 4.3

### Brief summary

Custom-persist can cause TemplateVM's private volume to be used as AppVMs' private volumes. This requires a very weird and specific set-up, but could cause information leakage if for some reason someone stores things in TemplateVMs' private volumes.

### Steps to reproduce

1. Enable custom-persist service on an AppVM
2. Set some directories to be persisted via `qvm-features` (I don't know whether that's important)
3. Enable custom-persist service on a TemplateVM
4. DO NOT set any directories to be persisted via `qvm-features`
5. Shut both VMs down
6. Launch the TemplateVM, perform some modifications in `/home/user`
7. Shut down the TemplateVM
8. Launch the AppVM 

### Expected behavior

Clean `/home/user` with "standard" dirs and files from `/etc/skel` and XDG.

### Actual behavior

Modifications made in the template in `/home/user` appear in the AppVM!

### Additional information

Always reproducible on my system.

Related: https://forum.qubes-os.org/t/any-drawbacks-to-enabling-custom-persist-in-templatevms/42564/8

datenbar:
Fixes QubesOS/qubes-issues/issues/11054

When `custom-persist` is
enabled on a TemplateVM with no paths persisted, the script falls through to
`initialize_home "/home" unconditionally` with no mount in place, but a
TemplateVM's root filesystem isn't discarded, so those writes land
permanently on the root volume and propagate to every AppVM based on that
template.

This adds an `is_templatevm` check ahead of the existing
`is_custom_persist_enabled` branch, so TemplateVMs always get the default
`mount_home=true` / `mount_usr_local=true` behavior regardless of
custom-persist configuration. AppVM behavior is unchanged.

## Testing

Tested with a mocked harness (qubesdb calls, mount, initialize_home, and
related functions stubbed out) covering four cases:
- TemplateVM + custom-persist enabled, no paths persisted → now takes the
  default branch (previously took the buggy fallback)
- AppVM + custom-persist enabled, no paths persisted → unchanged, still
  takes the fallback branch
- AppVM + custom-persist enabled, `/home` explicitly persisted → unchanged
- TemplateVM + custom-persist disabled → unchanged

## AI Disclosure
Used Claude to write test harness and check the fix before checking it myself.


vm-systemd/mount-dirs.sh
```diff
@@ -10,7 +10,10 @@ set -e
 if [ -e /dev/xvdb ] ; then mount /rw ; fi
 /usr/lib/qubes/init/setup-rw.sh
 
-if is_custom_persist_enabled; then
+if is_templatevm; then
```
rustybird: Checking `is_fully_persistent` seems like a better fit, and catches StandaloneVMs too (where custom-persist can't work either, for the same reason).

Would also be good to log a warning to stderr here, cause the user has tried to configure something that's not going to happen.

rustybird (changes requested):

datenbar:
CI failed due to reasons unrelated to the PR

marmarek:
PipelineRetry

rustybird (approved):
Nit: the commit message is outdated, and there's an unnecessary merge commit.

But the diff LGTM (untested). 👍

marmarek (approved):
I can pick just the fixing commit, skipping the extra merge

marmarek:
Picked in https://github.com/QubesOS/qubes-core-agent-linux/commit/4738333496c6b689207d8274d0f3425e796b6197

marmarek:
@rustybird thanks for helping with review!

